PR #60, opened 2024-02-01
Add schema on structure and values

tobi:
using cue to create a deep schema for the .json files

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,3 +30,14 @@ jobs:
     - name: Run tests
       working-directory: ./dev
       run: bin/rake test
+
+    - name: Set up Cue
+      uses: cue-lang/setup-cue@v1.0.0
+      with:
+        version: 'v0.7.0'
+
+    - name: Verify generation output
+      working-directory: ./dev
+      run: |
+        bin/generate_dist
+        bin/verify_dist_files


.gitignore
```diff
@@ -1,4 +1,6 @@
-/shopify
+.byebug_history
+schema/attribute_data.cue
+
 /_site
 /docs/_data
-.byebug_history
+/shopify
```
Chris911, line 6: I don't see this folder in `main` or the `gh-pages` branch. Still needed?

dev/bin/vet_schema
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
```
Chris911, line 1: Nit: should the script be named `vet_dist_files` instead of `vet_schema`?

schema/categories_schema.cue
```diff
@@ -0,0 +1,41 @@
+package shopify
+
+#categoryId: =~"^gid://shopify/Taxonomy/Category/\\w+(-\\d+)*$"
```
Chris911, line 3: Nit: `\w+` will include underscore and digits. If we want to be stricter we can go with: `[a-z]{2}`. Also use for `prefix` in `verticals`.

schema/categories_schema.cue
```diff
@@ -0,0 +1,41 @@
...
+
+// Structure for categories, including recursive definitions for children, attributes, and ancestors
+#category: {
+	id:        string & #categoryId
+	name:      string
+	level:     int & >=0
+	full_name: string
+	parent_id: *null | string & #categoryId
+	attributes: [...#attribute_ref]
+	children: [...#category_ref] | *[]
```
Chris911, line 32: Not sure the `| *[]` is necessary as we always have a `children`/`ancestors` key. Validation passes locally without it.

schema/attributes_schema.cue
```diff
@@ -0,0 +1,15 @@
+package shopify
+
+// Structure for attributes, including recursive definitions for values
+#attribute: {
+	id:   =~"^gid://shopify/Taxonomy/Attribute/\\d+$"
+	name: string
+	values: [...{
+		// id:   "\(id)/\\d+" // CUE can't handle nested checks with regexes yet
```
Chris911, line 8: This works locally with `#AttributeGid` defined as a constant. `\()` is used for [interpolation](https://cuelang.org/docs/tutorials/tour/expressions/interpolation/).

```
package shopify

#AttributeGid: "gid://shopify/Taxonomy/Attribute/\\d+"

// Structure for attributes, including recursive definitions for values
#attribute: {
	id:   =~"^\(#AttributeGid)$"
	name: string
	values: [...{
		id:   =~"^\(#AttributeGid)/\\d+$"
		name: string
	}]
}

version: =~"^\\d+.\\d+.\\d+$"
attributes: [...#attribute]
```

.gitignore
elsom25: Ah this was here before, but looks to be from the early days. Will 🔥

PR #61, opened 2024-02-01
Generate a taxonomy.json file

elsom25:
- [x] Extract serializers
- [x] Model attribute
- [x] Update Category to use new Attribute model properly
- [x] Add a "taxonomy" format for `Serializers::Json`


dev/src/attribute.rb
```diff
@@ -0,0 +1,56 @@
+require_relative 'storage/memory'
+
+class Attribute
+  @@nodes = {}
```
Chris911, line 4: Is this used?

PR #62, opened 2024-02-02
Use tree structure to start enforcing ordered and consistent outputs

elsom25:
We were relying on source to determine order before, which is fragile.

Given we're building a tree, let's make it an ordered tree and lean on depth-first traversal for generation.


sdieter-shopify (approved):
yay! this is a great improvement!

PR #65, opened 2024-02-13
Replace our build environment with a makefile, and refactor our schema for validation.

Senjai:
New hotness

image

I originally just wanted to buff up our cue schema for validation, but I saw an opportunity to clean up how we handle the build here too.

While this is mostly source available at this point, I introduced a makefile for two reasons:

- So that non shopifolk can still actually use the commands we outline.
- Because Make fits what it is we're actually doing, which is creating files from other files, but only when they change

On that second point, there's now a make clean if you want to build things from scratch, but really `make` will just do the right thing based on what it sees is out of date. 

The current implementation uses sentinel files because bin/generate_dist and bin/generate_docs generate arbitrary files. In a perfect world you could just pick what you need to output, but it hopefully isn't that big of a deal.

To run the tests, run `make test` which will run the cue schema check as well as any ruby tests.

### Cue things

The main goal by consolidating the file here is that I want to make it very clear in one file what the structure should look like for users. The structure is defined in `schema.cue`. The validations here are included to ensure referential integrity, as well as uniqueness of names and GID's. What's really important is that the categories are referentially correct when referencing attributes or other categories.

Cue isn't the fastest at this, so someone may want to do this in ruby later, but it's nice to have it as part of the schema. Only the lookup variables are used in the schema to show we are in fact guarenteeing referential integrity.

### DIFF
--- a/schema/schema.cue
+++ b/schema/schema.cue
@@ -1,7 +1,6 @@
 package product_taxonomy
 
-#attribute_gid_regex:       "^gid://shopify/Taxonomy/Attribute/\\d+$"
-#attribute_value_gid_regex: "^gid://shopify/Taxonomy/Attribute/\\d+/\\d+$"
+#attribute_gid_regex: "^gid://shopify/Taxonomy/Attribute/\\d+$"
 
 #category_gid_pattern: "^gid://shopify/Taxonomy/Category/[a-zA-Z]{2}(-\\d+)*$"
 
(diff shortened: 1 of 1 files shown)


schema/schema.cue
```diff
@@ -0,0 +1,55 @@
...
+version!: string & =~"^\\d+.\\d+.\\d+$"
+
+// Present in attributes.json / attributes_data.cue
+attributes!: [
+	...{
+		id!:   string & =~#attribute_gid_regex
+		name!: string
+		values!: [
+			...{
+				id!:   string & =~#attribute_value_gid_regex
```
elsom25, line 25: Curious if there was a particular reason you dropped the capture we had before (or maybe that wasn't doing what I was expecting):

```
	id:   =~"^\(#AttributeGid)$"
	// ...
	values: [...{
		id:   =~"^\(#AttributeGid)/\\d+$"
		// ...
	}]

elsom25: Non-blocking and happy to add in in a followup PR. Just curious if there was something I wasn't thinking about here

elsom25 (approved):
Thanks for the attribute id validation fix

PR #66, opened 2024-02-14
Simplify source data

elsom25:
- Update `Category` to not read `level` or `parent` from source data
- Update docs generation to be based off `Taxonomy` instead of source data
- Simplify source data to only: `id`, `name`, `children_ids`, and `attribute_ids`

### DIFF
--- a/data/categories/aa_apparel_accessories.yml
+++ b/data/categories/aa_apparel_accessories.yml
@@ -1,9 +1,6 @@
 ---
-- fully_qualified_type: Apparel & Accessories
-  id: aa
+- id: aa
   name: Apparel & Accessories
-  parent_id: 
-  level: 1
   children_ids:
   - aa-1
   - aa-2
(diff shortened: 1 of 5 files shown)


dev/src/category.rb
```diff
@@ -95,7 +101,7 @@ def add(child)
 
     child.parent = self
     children << child
-    children.sort!
+    children.sort!.uniq!
```
Chris911, line 104: Do we have duplicates in the source data or a code path that leads to duplicates? Perhaps we should change `children`/`attributes` to be `Set`.

Chris911 (approved):

dev/src/category.rb
elsom25: I had them as set's originally, but Ruby doesn't have ordered Set's 😮‍💨

Source data _shouldn't_ have duplicates, and we do have verification of no duplicates on the other side, so this isn't strictly needed

PR #93, opened 2024-03-08
add search_index.json to fix search

danloomer:
Commit `search_index.json` to fix search functionality.


danloomer:
Should be generated by build step... This is not the best fix.

PR #97, opened 2024-03-09
fix search on non chrome browsers and refactor into proper modules

danloomer:
Non chrome browsers don't currently support assert { type: 'json' } import guards, so the recently merged search feature broke the site for those browsers. This PR remedies that by switching to fetch to retrieve the search index and doing the necessary refactors to account for the async search index.

### DIFF
--- a/docs/assets/js/search.js
+++ b/docs/assets/js/search.js
@@ -81,7 +81,10 @@ async function categorySearch(query) {
   const fuse = await getFuse();
   const results = fuse.search(query, { limit: 5 });
   if (results.length === 0) {
-    searchContainer.textContent = "No results found";
+    const noResults = document.createElement("li");
+    noResults.textContent = "No results found";
+    noResults.style.minWidth = "160px";
+    searchContainer.appendChild(noResults);
     return;
   }
 


docs/assets/js/nodes.js
```diff
@@ -0,0 +1,144 @@
...
+  selectedNodes[depth] = initialNode;
+  selectedNode = initialNode;
+};
+
+export const resetToCategory = (categoryId) => {
+  selectedNodes = {};
+  setNodeQueryParam(categoryId);
+  setInitialNode();
+  renderPage();
+};
```
danloomer, line 138: this module is a straight copy of what was in index.js before related to nodes, except for this one function that I added which is the little bit of logic that search.js needs.

docs/assets/js/search.js
```diff
@@ -0,0 +1,102 @@
...
+    );
+    return searchIndex;
+  };
+})();
+
+const searchDebounceMs = 200;
+
+export const setupSearch = async () => {
+  // wait for fuse/searchIndex to load before setting up listeners
+  await getFuse();
```
danloomer, line 40: we probably should look into a nicer loading state but for now I changed the placeholder on load to "Loading" and then here we wait for fuse to be ready (+ search index internally), before updating the placeholder and adding listeners.

elsom25 (approved):

PR #100, opened 2024-03-11
better keyboard interactions:

danloomer:
- give clickable elements tabindex so that you can move around with the tab key
- attach event keypress event listener to handle enter key/space key for keyboard click
- secret bonus: added css variables for repetitive values, and used them in input to standardize border style

### DIFF
--- a/docs/assets/js/search.js
+++ b/docs/assets/js/search.js
@@ -42,7 +42,7 @@ export const setupSearch = async () => {
   const searchInput = q("#category-search");
   const searchResults = q("#category-search-results");
 
-  searchInput.placeholder = "Search for categories…";
+  searchInput.placeholder = "Search for categories";
 
   let timeoutId;
   searchInput.addEventListener("input", (e) => {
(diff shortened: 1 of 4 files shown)


docs/assets/js/search.js
```diff
@@ -42,7 +42,7 @@ export const setupSearch = async () => {
   const searchInput = q("#category-search");
   const searchResults = q("#category-search-results");
 
-  searchInput.placeholder = "Search for categories…";
+  searchInput.placeholder = "Search for categories";
```
danloomer, line 45: small editorial change, not sure why we had an ellipsis here 🤔

docs/styles.css
```diff
@@ -123,21 +131,29 @@ a:active {
...
   padding: 0;
-  border: 1px solid #ddd;
+  padding-bottom: 0.25rem;
+  border: var(--border);
+  box-shadow: var(--box-shadow);
 }
 
-#category-search-results li {
-  padding: 2px 4px;
+.category-search-container li {
```
danloomer, line 139: not really important for an app this small, but probably better practice to attach styles to a classname rather than an ID 🤷

elsom25 (approved):
💅🏼

PR #113, opened 2024-03-21
Add minitest hooks and make seeding fast

elsom25:
More robust testing, but more importantly **changes seeding from ~50s to ~2s**

### DIFF
--- a/app/serializers/data/category_serializer.rb
+++ b/app/serializers/data/category_serializer.rb
@@ -3,6 +3,10 @@
 module Serializers
   module Data
     class CategorySerializer < ObjectSerializer
+      class << self
+        delegate(:deserialize_for_insert_all, :deserialize_for_join_insert_all, to: :instance)
+      end
+
       def serialize(category)
         {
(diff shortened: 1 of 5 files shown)


app/models/category.rb
```diff
@@ -7,7 +7,14 @@ class Category < ApplicationRecord
...
   belongs_to :parent, class_name: "Category", optional: true
-  has_and_belongs_to_many :properties
+
+  has_and_belongs_to_many :properties,
+    join_table: :categories_properties,
+    foreign_key: :category_id,
+    association_foreign_key: :property_friendly_id
+  def property_friendly_ids=(ids)
+    self.properties = Property.where(friendly_id: ids)
+  end
```
KComrade53, line 17: I don't think this relationship is set up correctly. When I tried running it on your branch, the attributes array is empty for all the categories in categories.json

KComrade53: I imagine the other relationships will have similar issue, that's just the one I checked

PR #121, opened 2024-03-22
Add taxonomy mapping models and the logic to generate mappings from a list of mapping rules

chesterbot01:
### Description
- This PR introduces models for representing taxonomy mappings. `MappingRule` is defined as an input and output. Input and output are of `Product` type, which mean they are shaped like the products of the taxonomies being mapped.
- To generate mappings from a list of MappingRules, some services are introduced:
  - `InputProductGenerator` is used to generate combinations for all characteristic product categories and their attributes that appear in the input.
  - `MappingBuilder` is a compiler then passes each input through the set of matching rules, reducing the output of each rule into a final output
- Add [from_shopify_to_google taxonomy mapping rules](https://github.com/Shopify/product-taxonomy/blob/0b01e95275e3b8a7621468557f17b14ca9506eff/data/integrations/google/mappings/from_shopify.yml) in YAML.

### DIFF
--- a/app/models/google_product.rb
+++ b/app/models/google_product.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+class GoogleProduct < Product
+  store :payload, coder: JSON, accessors: [:product_category_id]
+end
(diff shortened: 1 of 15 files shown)


elsom25 (approved):
Before merging, can you post a copy of the local taxonomy viewer running to validate data generation continues to work as we expect? (aka run `make serve`)

Exciting step!

PR #123, opened 2024-03-26
Generate inline taxonomy mapping in taxonomy.json

chesterbot01:
Generate inline taxonomy mapping:
Screenshot 2024-03-25 at 10 17 11 PM

- [x] Taxonomy viewer works as normal

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ RUN_CMD  = printf "\e[%sm>> %-21s\e[0;1m\n" "1;34" # bold blue text with a >> pr
 CATEGORIES_DATA = $(shell find data/categories)
 ATTRIBUTES_DATA = $(shell find data/attributes.yml)
 VALUES_DATA     = $(shell find data/values.yml)
-MAPPINGS_DATA   = $(shell find data/integrations/*/mappings)
+MAPPINGS_DATA   = $(shell find data/integrations/*/*/mappings)
 
 ###############################################################################
 # TARGETS
--- a/app/serializers/dist/json_serializer.rb
+++ b/app/serializers/dist/json_serializer.rb
@@ -81,22 +81,24 @@ module Dist
     end
 
     def build_mapping_blocks
-      mapping_rule_blocks = Integration.all.flat_map do |integration|
-        [true, false].filter_map do |from_shopify|
-          rules = @mapping_rules.where(integration:, from_shopify:)
-          rules if rules.any?
+      mapping_rule_blocks = Integration.all.pluck(:id, :available_versions).flat_map do |id, versions|
+        [id].product(versions, [true, false])
+      end.filter_map do |integration_id, version, from_shopify|
+        rules = if from_shopify
+          @mapping_rules.where(integration_id:, from_shopify: true, output_version: version)
+        else
+          @mapping_rules.where(integration_id:, from_shopify: false, input_version: version)
         end
+        rules if rules.any?
       end
 
       mapping_blocks = mapping_rule_blocks.map do |mapping_rules|
         rules = Category.verticals.flat_map do |vertical|
           MappingBuilder.build_category_to_category_mappings_for_vertical(mapping_rules:, vertical:)
         end
-        from_shopify = mapping_rules.first.from_shopify
-        integration = mapping_rules.first.integration
         {
-          input_taxonomy: from_shopify ? "shopify/v1" : "#{integration.name}/v1",
-          output_taxonomy: from_shopify ? "#{integration.name}/v1" : "shopify/v1",
+          input_taxonomy: mapping_rules.first.input_version,
+          output_taxonomy: mapping_rules.first.output_version,
           rules:,
         }
       end
--- a/app/models/integration.rb
+++ b/app/models/integration.rb
@@ -3,4 +3,6 @@
 class Integration < ApplicationRecord
   has_many :mapping_rules
   validates :name, presence: true
+
+  serialize :available_versions, type: Array
 end
--- a/bin/seed
+++ b/bin/seed
@@ -16,7 +16,7 @@ attributes_data = cli.parse_yaml("data/attributes.yml")
 category_files = Dir.glob("#{CLI.root}/data/categories/*.yml")
 verticals_data = category_files.map { cli.parse_yaml(_1) }
 integrations_data = cli.parse_yaml("data/integrations/integrations.yml")
-mapping_rule_files = Dir.glob("#{CLI.root}/data/integrations/*/mappings/*_shopify.yml")
+mapping_rule_files = Dir.glob("#{CLI.root}/data/integrations/*/*/mappings/*_shopify.yml")
 
 Application.establish_db_connection!
 Application.load_and_reset_schema!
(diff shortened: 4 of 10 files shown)


app/serializers/dist/json_serializer.rb
```diff
@@ -70,5 +72,61 @@ def serialize_nested(connection)
         name: connection.name,
       }
     end
+
+    def build_mapping_blocks(mapping_rules)
```
elsom25, line 76: Use the instance variable instead of passing this in? The other ones take a param because they're operating against a subset of the instance var vs its entirety

app/serializers/dist/json_serializer.rb
```diff
@@ -70,5 +72,61 @@ def serialize_nested(connection)
...
+      all_mapping_blocks = []
+      puts "Generating mappings ..."
+      mapping_count = 0
+      Integration.all.each do |integration|
+        [true, false].each do |from_shopify|
```
elsom25, line 81:
```suggestion
      mapping_rules = Integration.all.flat_map do |integration|
        [true, false].filter_map do |from_shopify|
          rules = mapping_rules.where(integration:, from_shopify:)
          rules if rule.any?
        end
      end
```

elsom25: Mostly, if you decompose this it'll be easier to read with each piece being more focused. Also makes it simpler to use map methods, which are nice that each with manual tracking arrays 😁

app/serializers/dist/json_serializer.rb
```diff
@@ -70,5 +72,61 @@ def serialize_nested(connection)
...
+          next if mapping_rule_block.count.zero?
+
+          mappings = []
+          Category.verticals.each do |vertical|
+            mappings << MappingBuilder.build_one_to_one_mappings_for_vertical(
+              mapping_rules: mapping_rule_block,
+              vertical: vertical,
+            )
+          end
+          processed_mappings = mappings.flatten.compact
```
elsom25, line 95: AFAICT `compact` isn't needed here

```suggestion
          mappings = Category.verticals.flat_map do |vertical|
            MappingBuilder.build_one_to_one_mappings_for_vertical(mapping_rules:, vertical:)
          end
```

app/serializers/dist/json_serializer.rb
```diff
@@ -70,5 +72,61 @@ def serialize_nested(connection)
...
+      }
+    end
+
+    def serialize_mapping(mapping)
+      mapping[:input][:product_category_id] = Category.find(mapping[:input][:product_category_id]).gid
+      if mapping[:input][:attributes].present?
+        mapping[:input][:attributes] = mapping[:input][:attributes].map do |attribute|
+          {
+            name: Property.find(attribute[:name]).gid,
+            value: attribute[:value].nil? ? nil : PropertyValue.find(attribute[:value]).gid,
```
elsom25, line 122:
```suggestion
            value: attribute[:value] || PropertyValue.find(attribute[:value]).gid,
```

app/services/mapping_builder.rb
```diff
@@ -2,6 +2,16 @@
...
 class MappingBuilder
   class << self
+    def build_one_to_one_mappings_for_vertical(mapping_rules:, vertical:)
+      relevant_rules = mapping_rules.select { |rule| rule.input.product_category_id.start_with?(vertical.id) }
+      relevant_rules.map do |rule|
+        {
+          input: rule.input.payload.delete_if { |_k, v| v.nil? },
+          output: rule.output.payload.delete_if { |_k, v| v.nil? },
+        }
+      end
```
elsom25, line 12:
```suggestion
      mapping_rules
        .select { _1.input.product_category_id.start_with?(vertical.id) }
        .map do 
          {
            input: _1.input.payload.delete_if { |_k, v| v.nil? },
            output: _1.output.payload.delete_if { |_k, v| v.nil? },
          }
        end
```

app/serializers/dist/json_serializer.rb
```diff
@@ -70,5 +72,61 @@ def serialize_nested(connection)
...
+          end
+          processed_mappings = mappings.flatten.compact
+          mapping_count += processed_mappings.count
+          all_mapping_blocks << {
+            input_taxonomy: "shopify/v1",
+            output_taxonomy: "#{integration.name}/v1",
+            rules: processed_mappings,
+          }
+        end
+      end
```
elsom25, line 103: I had done this piece by piece, but assembling it:

```suggestion
      mapping_rules = Integration.all.flat_map do |integration|
        [true, false].filter_map do |from_shopify|
          rules = mapping_rules.where(integration:, from_shopify:)
          rules if rule.any?
        end
      end
      
      rules = Category.verticals.flat_map do |vertical|
        MappingBuilder.build_one_to_one_mappings_for_vertical(mapping_rules:, vertical:)
      end
      
      mapping_rules.map do |mapping|
        {
          input_taxonomy: "shopify/v1", # isn't this determined by the true/false technically?
          output_taxonomy: "#{mapping.integration.name}/v1",
          rules:,
        } 
      end
```

chesterbot01: Thanks! Simplified the code based on the suggestion.

app/serializers/dist/json_serializer.rb
```diff
@@ -2,9 +2,10 @@
 
 module Dist
   class JSONSerializer
-    def initialize(verticals:, properties:, version:)
+    def initialize(verticals:, properties:, mapping_rules: nil, version:)
```
elsom25, line 5: Do we need default value here?

```suggestion
    def initialize(verticals:, properties:, mapping_rules:, version:)
```

app/serializers/dist/json_serializer.rb
```diff
@@ -70,5 +79,56 @@ def serialize_nested(connection)
         name: connection.name,
       }
     end
+
+    def build_mapping_blocks
+      puts "Generating mappings ..."
```
elsom25, line 84: Do we need this in our serializers? I think it's reasonable for serializers to not have console output 😅

app/serializers/dist/json_serializer.rb
```diff
@@ -70,5 +79,54 @@ def serialize_nested(connection)
         name: connection.name,
       }
     end
+
+    def build_mapping_blocks
+      mapping_count = 0
```
elsom25, line 84: We don't need this anymore, correct?

app/serializers/dist/json_serializer.rb
chesterbot01: I have to revert the change here because in some cases a nil attribute value will lead to a specific output that has been defined in YAML mapping rules.

elsom25: Ah fair. Kinda makes me want a `NilPropertyValue` instead of magic nil, but this is fine

Makefile
```diff
@@ -52,6 +54,7 @@ LOCAL_DB = tmp/local.sqlite3
 # CUE imports needed for schema validation
 ATTRIBUTES_DATA_CUE = schema/attributes_data.cue
 CATEGORIES_DATA_CUE = schema/categories_data.cue
+MAPPINGS_DATA_CUE = schema/mappings_data.cue
```
elsom25, line 57: You only need this if you need the data generated as part of integrity checks (see `validations.cue`)

chesterbot01: without `schema/mappings_data.cue`, `cue vet` will error out 
```
mappings: field is required but not present:
    ./schema.cue:59:1
```

PR #125, opened 2024-04-02
More explicit relationship between values and properties

elsom25:
- [x] Add Value#primary_property
- [x] Change txt serializations to leverage this
- [x] Update `attributes.yml` to have normative IDs

As for output for `attribute_values.txt`, I went with:
- `# Format: {GID} : {Name} [{Attribute Name}]`

Considered these as well. `>` was tempting but leads to hard-to-read files with mis-alignment:
- `# Format: {GID} : {Attribute name} > {Value name}`
- `# Format: {GID} : {Name} [{All attribute names separated by ;}]`

<details><summary>Wood with primary only</summary>
<pre>
gid://shopify/TaxonomyValue/67    : Wood [Material]
gid://shopify/TaxonomyValue/309   : Wood [Decoration material]
gid://shopify/TaxonomyValue/490   : Wood [Door material]
gid://shopify/TaxonomyValue/521   : Wood [Food storage material]
gid://shopify/TaxonomyValue/569   : Wood [Habitat material]
gid://shopify/TaxonomyValue/889   : Wood [Clothing accessory material]
gid://shopify/TaxonomyValue/958   : Wood [Jewelry material]
gid://shopify/TaxonomyValue/978   : Wood [Bristle material]
gid://shopify/TaxonomyValue/1491  : Wood [Fastener material]
gid://shopify/TaxonomyValue/1551  : Wood [Suitable for crafting material]
gid://shopify/TaxonomyValue/1628  : Wood [Wreath design]
gid://shopify/TaxonomyValue/2067  : Wood [Blade material]
gid://shopify/TaxonomyValue/2097  : Wood [Handle material]
gid://shopify/TaxonomyValue/2220  : Wood [Brush material]
gid://shopify/TaxonomyValue/2750  : Wood [Scale model material]
gid://shopify/TaxonomyValue/2906  : Wood [Instrument accessory material]
gid://shopify/TaxonomyValue/3134  : Wood [Drum hardware material]
gid://shopify/TaxonomyValue/3150  : Wood [Mount material]
gid://shopify/TaxonomyValue/3198  : Wood [Mallet material]
gid://shopify/TaxonomyValue/3252  : Wood [Instrument part material]
gid://shopify/TaxonomyValue/3392  : Wood [Instrument material]
gid://shopify/TaxonomyValue/3636  : Wood [Accessory material]
gid://shopify/TaxonomyValue/4301  : Wood [Shelf material]
gid://shopify/TaxonomyValue/7966  : Wood [Compatible filament]
[...]

PR #129, opened 2024-04-05
Add test to make sure attributes aren't missing values

KComrade53:
Attributes should either have `values_from` when they inherit values, or `values` if they have their own list of values. In the latter case, the values array should not be empty.

This is a followup to: https://github.com/Shopify/product-taxonomy/pull/125#discussion_r1551917060

### DIFF
--- a/test/integration/all_data_files_import_test.rb
+++ b/test/integration/all_data_files_import_test.rb
@@ -130,6 +130,16 @@ class AllDataFilesImportTest < ActiveSupport::TestCase
     end
   end
 
+  test "Attributes have at least one value listed or inherited" do
+    @raw_attributes_data.each do |raw_attribute|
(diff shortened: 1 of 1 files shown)


test/integration/all_data_files_import_test.rb
```diff
@@ -130,6 +130,16 @@ def before_all
     end
   end
 
+  test "Attributes have at least one value listed or inherited" do
```
elsom25, line 133: This test is already covered at the top with the `xor` test

elsom25:
https://github.com/Shopify/product-taxonomy/blob/main/test/integration/all_data_files_import_test.rb#L28-L38 should cover it.

And when we add dimensions, I suspect we ad a "third" type: `value_validations` or something to that effect?

PR #130, opened 2024-04-05
Add categories for Gift Cards and Product Add-ons

KComrade53:
Add categories for both the Gift Cards and Product Add-ons taxonomies


elsom25:
Let's add a changelog and do a version bump? (Can't use the script yet since that's meant for _real releases_ and we're not there yet)

KComrade53:
@elsom25 good call, added

PR #134, opened 2024-04-12
Simplify mapping compilation

chesterbot01:
This PR simplifies the generation of mappings. 
There will be a followup PR to clean up methods that are no longer used.

### DIFF
--- a/app/serializers/source_data/product_serializer.rb
+++ b/app/serializers/source_data/product_serializer.rb
@@ -18,7 +18,7 @@ module SourceData
       if product_type == "ShopifyProduct"
         properties = payload.delete("attributes")
         payload["properties"] = properties
-        payload["product_category_id"] = "gid://shopify/TaxonomyCategory/#{payload["product_category_id"]}"
+        payload["product_category_id"] = Category.gid(payload["product_category_id"])
       end
       Product.new(payload: payload, type: product_type)
     end
--- a/app/models/category.rb
+++ b/app/models/category.rb
@@ -25,8 +25,14 @@ class Category < ApplicationRecord
     unless: :root?
   validates_associated :children
 
+  class << self
+    def gid(id)
+      "gid://shopify/TaxonomyCategory/#{id}"
+    end
+  end
+
   def gid
-    "gid://shopify/TaxonomyCategory/#{id}"
+    self.class.gid(id)
   end
 
   def full_name


data/integrations/integrations.yml
```diff
@@ -2,3 +2,6 @@
 - name: google
   available_versions:
     - google/2021-09-21
+- name: shopify
+  available_versions:
+    - shopify/2022-02
```
jeanrick, line 7: We should leave this entry for https://github.com/Shopify/product-taxonomy/pull/127 which adds this mapping

schema/schema.cue
```diff
@@ -62,7 +62,7 @@ mappings!: [
 		rules!: [
 			...{
 				input!: {
-					product_category_id: string & =~#category_gid_regex
+					product_category_id: string
```
jeanrick, line 65: This is fine for now until we add a means for the schema to be defined for each input/output type 🤔

app/serializers/source_data/product_serializer.rb
```diff
@@ -18,6 +18,7 @@ def deserialize(payload, product_type)
       if product_type == "ShopifyProduct"
         properties = payload.delete("attributes")
         payload["properties"] = properties
+        payload["product_category_id"] = "gid://shopify/TaxonomyCategory/#{payload["product_category_id"]}"
```
elsom25, line 21: Any reason this can't come from `Category#gid`?

jeanrick, line 21: _nit_ Would be nice to have a reusable method for this. Like `Category.gid`

Does this gid work for the old taxonomy @elsom25?

jeanrick (approved):

app/serializers/source_data/product_serializer.rb
chesterbot01: Shopify/2022-02's category nodes are not seeded

app/services/mapping_builder.rb
```diff
@@ -2,6 +2,16 @@
 
 class MappingBuilder
   class << self
+    def simple_mapping(mapping_rules:)
+      mapping_rules
+        .map do
+          {
+            input: _1.input.payload.delete_if { |_k, v| v.nil? },
+            output: _1.output.payload.delete_if { |_k, v| v.nil? },
```
elsom25, line 10:
```suggestion
            input: _1.input.payload.compact,
            output: _1.output.payload.compact,
```

elsom25 (approved):
Two small tune-ups but broadly okay for now. Thanks for getting this together quickly

app/serializers/source_data/product_serializer.rb
jeanrick: Ah, I see there's already a similar comment thread: https://github.com/Shopify/product-taxonomy/pull/134/files#r1563212184

data/integrations/integrations.yml
chesterbot01: removed.

PR #137, opened 2024-04-23
Introduce "Extended" attributes

elsom25:
Extended properties allow for better names in different categories, while ensuring values are consistent.

- [x] Update models
- [x] Update serializers
- [x] Remove data validation from cue and lean on our DB+test suite instead
- [ ] Split base and extended properties in data files


app/models/property.rb
```diff
@@ -22,6 +22,23 @@ def property_value_friendly_ids=(ids)
   validates :name, presence: true
 
   def gid
-    "gid://shopify/TaxonomyAttribute/#{id}"
+    if abstract?
```
elsom25, line 25: Kinda inclined to just have two models... maybe STI?

elsom25:
Replaced by #139

PR #139, opened 2024-04-24
Add Handles to Attribute/Value

tylerrowsell:
The output of the changes in https://github.com/Shopify/product-taxonomy-tools/pull/20

### DIFF
--- a/schema/schema.cue
+++ b/schema/schema.cue
@@ -14,17 +14,20 @@ version!: string & =~"^\\d+.\\d+.\\d+$"
 // Present in attributes.json / taxonomy.json
 attributes!: [
 	...{
-		id!:   string & =~#attribute_gid_regex
-		name!: string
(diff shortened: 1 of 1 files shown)


elsom25 (approved):

application.rb
```diff
@@ -4,6 +4,7 @@
 require "fileutils"
 require "json"
 require "yaml"
+require "byebug"
```
elsom25, line 7: Shouldn't ever need this (we autoload all gems right below this)

app/models/property.rb
```diff
@@ -20,8 +33,31 @@ def property_value_friendly_ids=(ids)
   end
 
   validates :name, presence: true
+  validates :friendly_id, presence: true, uniqueness: true
+  validate :property_values_match_base, if: :extended?
+  validates :handle, presence: true
```
elsom25, line 38: Let's put this one above the `validate`?

PR #143, opened 2024-04-30
April 30th Update

tylerrowsell:
Importing the latest updates from the CSV files.

Includes https://github.com/Shopify/custom-data-issues/issues/2609

### DIFF
--- a/data/localizations/values/cs.yml
+++ b/data/localizations/values/cs.yml
@@ -7227,6 +7227,8 @@ cs:
       name: MMC micro
     compatible_memory_cards__mmc_mobile:
       name: MMC mobile
+    compatible_memory_cards__mmcplus:
+      name:
     compatible_memory_cards__ms:
       name: MS
     compatible_memory_cards__ms_duo:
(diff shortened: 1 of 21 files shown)


data/localizations/categories/cs.yml
```diff
@@ -895,6 +895,8 @@ cs:
       name: Obuv
     aa-8-1:
       name: Atletická obuv
+    aa-8-10:
```
elsom25, line 898: the sort order 😢

elsom25 (approved):

data/localizations/categories/cs.yml
KComrade53: @tylerrowsell if you split the string by "-" then it can sort the resultant arrays so that [aa,8,2] will be before [aa,8,10]. I did so in the parser overriding the `<=>` method with that logic.

Feel free to poke me on slack if you want help or have questions

tylerrowsell: oh gross. Didn't even notice this. I'll push a fix before merging.

PR #157, opened 2024-05-10
Move MappingSerializer into MappingRule for now

elsom25:
Mostly moving the code as-is out of the serializer so we can delete that class. Due for a cleanup, but shouldn't gate.


alexmgrant (approved):
👍

PR #160, opened 2024-05-13
Localization Dist

tylerrowsell:
Creates a new folder structure for dist

- Dists will be generated per languages i.e. `dist/en`, `dist/fr` etc.
- Adds the plumbing to generate dists in all locales
- Adds a new github action to generate non english dists on commit to main.


app/models/category.rb
```diff
@@ -124,8 +141,12 @@ def gid
...
   end
 
-  def full_name
-    ancestors.reverse.map(&:name).push(name).join(" > ")
+  def name(locale: "en")
+    self.class.find_localization(locale, id, "name") || super()
+  end
+
+  def full_name(locale: "en")
+    ancestors.map { _1.name(locale:) }.push(name(locale:)).join(" > ")
```
elsom25, line 149: Lost the `reverse` here
```suggestion
    ancestors.reverse.map { _1.name(locale:) }.push(name(locale:)).join(" > ")
```

elsom25 (changes requested):
A few changes I think we need, but this is looking _Way better_

bin/generate_dist
```diff
@@ -4,76 +4,63 @@
...
+end
+@cli.parse!(ARGV)
+@cli.options.version ||= @cli.read_file("VERSION").strip
 
-      locales.each do |locale|
-        generate_dist_files(locale)
-      end
-    end
+@cli.options.locales ||= ["en"]
+if @cli.options.locales.include?("ALL")
```
elsom25, line 15: small aside: I'm unsure if OptionParser downcases for us or not (just something to vet locally)

elsom25 (approved):
Thanks for the changes — this looks awesome! 🚀

PR #172, opened 2024-05-22
Adopt rails proper

elsom25:
Just use rails fully. Not needed _yet_ (though does improve interactions of Ruby LSP), but key for work on internal tooling coming soon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,10 @@ Everything else is how we manage the taxonomy and generate distributions. This i
 
 ## 🤿 Diving in
 
-This is a simple ruby app with a few models and serializers. The bulk of the work is parsing `data/` into a tree of `app/models/category.rb` to serialize reliably to `dist/`. The app is set up to be rails-like including using `ActiveRecord`.
+This is a simple ETL app ultimately, with a few core models. The app is built on Ruby, Rails, and Jekyll:
+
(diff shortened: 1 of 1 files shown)


tylerrowsell (approved):
Let's do it!

.gitignore
```diff
@@ -5,7 +5,7 @@ schema/mappings_data.cue
 
 # Ignore docs site
 /_site
-/docs/_data/*
+/docs/_data/unstable
```
elsom25, line 8: We want unstable ignored, but we want all published versions snapshotted

test/factories/mapping_rule_factory.rb
```diff
@@ -2,8 +2,10 @@
 
 FactoryBot.define do
   factory :mapping_rule do
-    sequence(:id, 1)
-    sequence(:integration_id, 1)
+    association :integration, factory: :integration
+    association :input, factory: :product
+    association :output, factory: :product # TODO: FactoryBot is upset about this
```
elsom25, line 7: @alexmgrant I opted to skip the factory validation tests for now, but FactoryBot was unhappy with this.

I figure since y'all are reworking things skipping was fine, but let's try and make sure whatever comes next gets our factory validation test back on

PR #178, opened 2024-05-28
Remove Checkout from Distribution Flow

tylerrowsell:
Remove Checkout from Dist Flow

Updates to make a PR instead using https://github.com/peter-evans/create-pull-request/tree/v6/

PR #179, opened 2024-05-28
Bugfix: Adjust async localization generation Github Action

elsom25:
I _think_ this is everything to tune here... may try and make the commit message a touch nicer (e.g., reference the merge commit for example)


elsom25:
Ah https://github.com/Shopify/product-taxonomy/pull/178 covers it

PR #199, opened 2024-05-30
add full names for dist/json mapping files

kott:
This PR adds `full_name` to the distribution JSON files for taxonomy mappings. Along with this, it stores the mappings in the new directory structure:
* `all_mappings` are now under `dist/en/integrations/all_mappings.json`
* each integration now has its own mappings file, e.g. `dist/en/integrations/shopify/shopify_2022-02_to_shopify_2024-04.json`

### DIFF
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -16,6 +16,10 @@ class Product < ApplicationRecord
       find_or_create_by!(type:, payload: payload_for(data, type:), full_name:)
     end
 
+    def full_name_map
+      @full_name_map ||= {}
+    end
+
     private
 
     def payload_for(data, type:)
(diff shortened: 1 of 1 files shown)


app/models/mapping_rule.rb
```diff
@@ -52,16 +52,25 @@ def mapping_blocks_as_json(input_version, output_version, mapping_rules)
...
 
   def as_json
-    resolve_input_attribute_values!
+    input_integration_version = input_version unless from_shopify?
+    output_integration_version = output_version if from_shopify?
+
+    input_json = input.as_json(integration_version: input_integration_version)
+    output_json = output.as_json(integration_version: output_integration_version)
+
+    if input_json.empty? || output_json.empty?
```
kott, line 67: This is a temporary solution to avoid writing mappings that do not have a corresponding category in shopify. There are currently some categories in the mappings that no longer exist in the shopify categories. In the future we will have checks for this scenario before writing to the data files, but as of now we have this scenario that we need to code against.

app/models/product.rb
```diff
@@ -57,11 +57,63 @@ def as_json_for_data
...
+      }
+    end
+  end
+
+  def integration_full_name(category_id:, integration_version:)
+    full_name_map(integration_version:)[category_id.to_s]
+  end
+
+  def full_name_map(integration_version:)
+    categories = YAML.load_file(File.join(Rails.root, "data/integrations/#{integration_version}/full_names.yml"))
```
Fernanda-Montiel, line 114: It's better to have the YAML loaded once, instead of each iteration (It happened to me on the pipeline and the difference on the time of processing was huge)

kott: Good catch! I've memoized it on a per integration basis now at the class level, so it should be a little more speedy.

Fernanda-Montiel (approved):

PR #206, opened 2024-06-05
Checking category IDs in mappings are valid

Catherine-Fu:
Checking categories IDs used in mappings are present in corresponding category files

- add this consistency validation in an integration testing 
- this validation takes 0.29s 

Testing result: 

test_result

### DIFF
--- a/test/integration/mapping_validation_test.rb
+++ b/test/integration/mapping_validation_test.rb
@@ -3,7 +3,7 @@
 require_relative "../test_helper"
 require_relative "../../config/cli"
 
-class LocalizationsTest < ActiveSupport::TestCase
+class MappingValidationTest < ActiveSupport::TestCase
   test "category IDs in mappings are valid" do
     mappings_json_data = CLI.new.parse_json("dist/en/integrations/all_mappings.json")
     invalid_categories = []


test/integration/mapping_validation_test.rb
```diff
@@ -0,0 +1,110 @@
...
+require_relative "../test_helper"
+require_relative "../../config/cli"
+
+class LocalizationsTest < ActiveSupport::TestCase
+  test "category IDs in mappings are valid" do
+    mappings_json_data = CLI.new.parse_json("dist/en/integrations/all_mappings.json")
+    invalid_categories = []
+    mappings_json_data["mappings"].each do |mapping|
+      # process input taxonomy
+      category_ids = get_category_ids(mapping["input_taxonomy"])
```
tylerrowsell, line 12: nit: I'd move this into `validate_mapping_category_ids` since you already have to pass in `mapping["input_taxonomy"]` there.

test/integration/mapping_validation_test.rb
```diff
@@ -0,0 +1,110 @@
...
+      }
+    else
+      {
+        all_mapping_category_ids_are_valid: false,
+        invalid_categories: invalid_categories,
+      }
+    end
+  end
+
+  def get_category_ids(input_or_output_taxonomy)
```
tylerrowsell, line 87: nit: Generally we don't use `get_` in ruby. `category_ids` or `category_ids_from_taxonomy`

test/integration/mapping_validation_test.rb
```diff
@@ -0,0 +1,110 @@
...
+    cli = CLI.new
+    if input_or_output_taxonomy.include?("shopify/2022-02")
+      nil
```
tylerrowsell, line 90: guard clause before initializing the CLI?
```suggestion
    return if input_or_output_taxonomy.include?("shopify/2022-02")
```

test/integration/mapping_validation_test.rb
tylerrowsell, line 87: nit: `taxonomy` instead of `input_or_output_taxonomy`

test/integration/mapping_validation_test.rb
```diff
@@ -0,0 +1,110 @@
...
+        invalid_categories.push(res[:invalid_categories])
+      end
+    end
+
+    assert invalid_categories.empty?,
+      "The following category ids in mappings are invalid:
+    #{invalid_categories}"
+  end
+
+  def validate_mapping_category_ids(mapping_rules, category_ids, input_or_output, input_or_output_taxonomy)
```
tylerrowsell, line 31: Seems like there might be some redundancy here. I believe the only difference between input/output is the category id being an item vs array. I threw this into chat gpt... but I think this would work and clean up the code? 

```
return { all_mapping_category_ids_are_valid: true } if category_ids.nil?

  invalid_category_ids = Set.new

  mapping_rules.each do |rule|
    product_categories = rule[input_or_output]["category"]
    product_categories = [product_categories] unless product_categories.is_a?(Array)
    
    product_categories.each do |product_category|
      invalid_category_ids.add(product_category["id"]) unless category_ids.include?(product_category["id"])
    end
  end

  if invalid_category_ids.empty?
    { all_mapping_category_ids_are_valid: true }
  else
    {
      all_mapping_category_ids_are_valid: false,
      invalid_categories: [
        {
          taxonomy_version: input_or_output_taxonomy,
          category_ids: invalid_category_ids,
        }
      ]
    }
  end
```

tylerrowsell (approved):
Looks like this test will do the trick. I added some nits/notes on how you might be able to cleanup the code to make it more ruby like and DRY it up a little bit.

test/integration/mapping_validation_test.rb
Catherine-Fu: Done

test/integration/mapping_validation_test.rb
Catherine-Fu: Done

test/integration/mapping_validation_test.rb
Catherine-Fu: Done

test/integration/mapping_validation_test.rb
Catherine-Fu: The reason of using the name `input_or_output_taxonomy` is to indicate this parameter expects the value of `mapping["input_taxonomy"]` or `mapping["output_taxonomy"]`.

chesterbot01 (approved):

PR #208, opened 2024-06-05
Amg/simple taxonomy mappings

alexmgrant:
closes: https://github.com/Shopify/temp-project-mover-Archetypically-20250318143024/issues/342

[Design](https://www.figma.com/design/dOTJttOyHa2ISyl35Lvsiq/Taxonomy-Jam?node-id=16-311&t=fCh7yGnXhw9H612F-0)

Adds taxonomy mappings to visualizer web app, improves performance and dev environment for Jekyll site by:

- moved to [BEM](https://getbem.com/) css pattern to simplify selectors for performance gains
  - [when toggling visibility](https://github.com/Shopify/product-taxonomy/pull/208/files#diff-d0c5f424f8ee0a09aa45e8f4ef7969a03f758a8d764b001bd634101b1064c13fR67) using JS selectors a `hidden` & `visible` class is used explicitly reducing style rendering times
  - removed wrapping "card" container around app as it was causing unnecessary DOM reflow events  
- [moved to standardized](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes) `data-*` attribute for JS consumption
- [caching DOM reads](https://github.com/Shopify/product-taxonomy/pull/208/files#diff-d0c5f424f8ee0a09aa45e8f4ef7969a03f758a8d764b001bd634101b1064c13fR10) after first paint. This reduces task length on the main thread
  - where possible DOM reads are done before writes. [ex](https://github.com/Shopify/product-taxonomy/pull/208/files#diff-d0c5f424f8ee0a09aa45e8f4ef7969a03f758a8d764b001bd634101b1064c13fR64)
-  `renderPage` uses either [yieldToMain](https://github.com/Shopify/product-taxonomy/pull/208/files#diff-d0c5f424f8ee0a09aa45e8f4ef7969a03f758a8d764b001bd634101b1064c13fR163) or [schedular](https://github.com/Shopify/product-taxonomy/pull/208/files#diff-d0c5f424f8ee0a09aa45e8f4ef7969a03f758a8d764b001bd634101b1064c13fR163) [when available](https://github.com/Shopify/product-taxonomy/pull/208/files#diff-d0c5f424f8ee0a09aa45e8f4ef7969a03f758a8d764b001bd634101b1064c13fR188) to break of tasks on main thread. The aim is 60fps, 16ms per task
- adding `eslint` and `prettier` to dev environment. I left it optional by not running `pnpm install` in `dev up`
  
## Testing the app:
- `dev up`
- `dev build`
- make sure the local Jekyll server is running from [the Procfile](https://github.com/Shopify/product-taxonomy/blob/017654152fea195e3ce220de455f0baa744257c0/.spin/Procfile#L1)
  - either run the server command explicitly `jekyll serve --source docs --port 2399` or within vscode Spin tools internally
- navigate to http://localhost:2399/

Sometimes in Spin (internally) the port stops forwarding to your local computer and visiting http://localhost:2399/ will not load anything. Deleting the port and re-adding it within the vscode PORTS tab will fix it.  
Screenshot 2024-06-06 at 1 17 36 PM

### DIFF
--- a/lib/docs/mappings.rb
+++ b/lib/docs/mappings.rb
@@ -1,8 +1,11 @@
+# frozen_string_literal: true
+
 module Docs
   class Mappings
     def reverse_shopify_mapping_rules(mappings)
       mappings.each do |mapping|
         if shopify_mapping?(mapping)
+          reverse_taxonomy_names(mapping)
           reverse_mapping_rules(mapping)
         end
       end
@@ -12,12 +15,20 @@ module Docs
       mapping["rules"].each do |rule|
         rule["output"]["category"].each_with_index do |output, index|
           if index == 0
-            rule["input"], rule["output"] = { "category" => output }, {"category" => [rule["input"]["category"]]}
+            rule["output"] = { "category" => [rule["input"]["category"]] }
+            rule["input"] = { "category" => output }
           else
             mapping.push = build_rule(input: output, output: rule["input"]["category"])
           end
         end
       end
+    end
+
+    def reverse_taxonomy_names(mapping)
+      input = mapping["input_taxonomy"]
+      output = mapping["output_taxonomy"]
+      mapping["input_taxonomy"] = output
+      mapping["output_taxonomy"] = input
     end
 
     def shopify_mapping?(mapping)
@@ -26,8 +37,8 @@ module Docs
 
     def build_rule(input:, output:)
       {
-        "input" => {"category" => input},
-        "output" => {"category" => [output]}
+        "input" => { "category" => input },
+        "output" => { "category" => [output] },
       }
     end
   end


docs/_config.yml
```diff
@@ -1,4 +1,7 @@
+exclude: [node_modules, README.md]
+incremental: true
```
alexmgrant, line 2: allows incremental build for faster local dev build times

docs/_layouts/release.html
```diff
@@ -85,7 +99,43 @@ <h4 class="title attribute-title" tabindex="0">
       {% endfor %}
     </div>
   </div>
+  <div class="column">
```
alexmgrant, line 102: new mappings

alexmgrant:
> This may be a bug in the mappings data, but we should be showing public IDs here... image

Odd that looks like an older version of the app...

kott (approved):

tylerrowsell (approved):

PR #211, opened 2024-06-06
Yaml to CSV converter

chesterbot01:
See description in the linked ticket.


tylerrowsell:
I think this would be better suited to be put in the https://github.com/shopify/product-taxonomy-tools repo.

PR #214, opened 2024-06-11
Amg/analytics

alexmgrant:
Adds Google Analytics to visualizer app.


docs/assets/js/nodes.js
```diff
@@ -184,16 +217,38 @@ const renderWithScheduler = () => {
   scheduler.postTask(toggleMappedCategories);
 };
```
alexmgrant, line 219: A bit robust to support `scheduler` when available. Keeps things feeling snappy by putting the analytics call in the background.

PR #225, opened 2024-07-02
Use gem to sort size-based attribute values

KComrade53:
Attribute values are currently being sorted alphabetically in the [Taxonomy Viewer](https://shopify.github.io/product-taxonomy/releases/unstable/?categoryId=gid%3A%2F%2Fshopify%2FTaxonomyCategory%2Faa-8), which doesn't make sense for all attributes, but also causes issues where the values are numerics. 

This change leverages Shopify's filter_values_sorter gem to clean up the values sorting for the taxonomy viewer so the values are displayed in the desired order, which matches the order they are displayed on the product editor of the Shopify site.

**Note**: This gem is shopify internal-only, so public contributors will not have access to this preferred sorting. In that case, the code will sort alphabetically as before, but then a Shopify employee can regenerate the dist files using the gem to fix the sorting before we make any new releases.

Resolves https://github.com/Shopify/custom-data-issues/issues/3127

### Before

Screenshot 2024-07-02 at 12 24 21 PM

### After

Screenshot 2024-07-02 at 11 15 15 AM
Screenshot 2024-07-02 at 11 15 31 AM

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,7 +22,7 @@ jobs:
       with:
         bundler-cache: true
       env:
-        BUNDLE_PKGS__SHOPIFY__IO: token:${{ secrets.CLOUDSMITH_RUBY_TOKEN_RO }}
+        BUNDLE_PKGS__SHOPIFY__IO: ${{ secrets.RUBYGEMS_SERVER_PKGS_SHOPIFY_IO_USERNAME }}:${{ secrets.RUBYGEMS_SERVER_PKGS_SHOPIFY_IO_PASSWORD }}
         BUNDLE_WITH: optional
     - uses: cue-lang/setup-cue@v1.0.0
       with:


tylerrowsell (commented):
1. Let's make sure to add some tests for ValuesSorter
2. Should we also sort on the values serializers? 

https://github.com/Shopify/product-taxonomy/blob/6b7bb7499a561d91c21919a7ee41840d9e032a64/app/models/value.rb#L51-L68

tylerrowsell:
Thinking out loud here.... do we actually want all the dist/jsons to be sorted here? Feels like they should be sorted by ID? Perhaps we should make it so this sort only happens when generating the docs pages?

tylerrowsell:
> I can look into updating the docs generation logic to just handle the sorting at the expense of making its logic more complex

I think that'd be okay. We have doc specific serializers on category. So we could do something similar on attributes.
https://github.com/Shopify/product-taxonomy/blob/6b7bb7499a561d91c21919a7ee41840d9e032a64/app/models/category.rb#L78-L119

I'm doing something similar in this PR anyways... so I'll rebase off of that.
https://github.com/Shopify/product-taxonomy/pull/226/files#diff-290fa384119e4dbcf3610128e28c3b0bc75b6b8ad64d6256b34720a47ec5796aR74

dev.yml
```diff
@@ -8,6 +8,10 @@ type: ruby
 
 up:
   - ruby
+  - custom:
+      name: "Include optional gems with bundle install"
+      met?: 'cat .bundle/config | grep "BUNDLE_WITH: \"optional\""'
+      meet: bundle config set --local with 'optional'
   - bundler
```
tylerrowsell, line 15: Is this bundler command still needed?

tylerrowsell (approved):

dev.yml
KComrade53: Yes, all the previous `custom` command is doing is adding a config to the bundler, which will be applied when it runs install

PR #226, opened 2024-07-02
Attribute docs page with search

tylerrowsell:


tylerrowsell:
> Not sure what the intent on this one is, but lets make sure we dont delete the 2024-07 docs static snapshot 😅

This was something I was hacking on with Ricardo. We're trying to add a "attributes" page.

i.e. Search for an attribute, see what values/categories it is attached to. 

No plans to merge without a lot of cleanup first.

PR #230, opened 2024-07-08
Sort attributes' values in docs by order specified in attributes data yml file

KComrade53:
This PR replaces https://github.com/Shopify/product-taxonomy/pull/225

Attribute values are currently being sorted alphabetically in the [Taxonomy Viewer](https://shopify.github.io/product-taxonomy/releases/unstable/?categoryId=gid%3A%2F%2Fshopify%2FTaxonomyCategory%2Faa-8), which doesn't make sense for all attributes, but also causes issues where the values are numerics. 

This PR updates the docs generation logic to display each attribute's values in the same order they are listed in the `data/attributes.yml` file.

Resolves https://github.com/Shopify/custom-data-issues/issues/3127

**Note**: Because of the large number of changed lines when applying this new sorting to the dist json files for each locale, I have split those changes into multiple PRs. Only English is included in this PR.

### Before

Screenshot 2024-07-02 at 12 24 21 PM

### After

Screenshot 2024-07-02 at 11 15 15 AM
Screenshot 2024-07-02 at 11 15 31 AM

### DIFF
--- a/dist/en/attributes.json
+++ b/dist/en/attributes.json
@@ -530,19 +530,9 @@
       ],
       "values": [
         {
-          "id": "gid://shopify/TaxonomyValue/11119",
-          "name": "Compact",
-          "handle": "accessory-size__compact"
-        },
-        {
-          "id": "gid://shopify/TaxonomyValue/15045",
-          "name": "Customizable",
-          "handle": "accessory-size__customizable"
-        },
-        {
-          "id": "gid://shopify/TaxonomyValue/2917",
-          "name": "Double extra large (XXL)",
-          "handle": "accessory-size__double-extra-large-xxl"
+          "id": "gid://shopify/TaxonomyValue/17114",
+          "name": "Triple extra small (XXXS)",
+          "handle": "accessory-size__triple-extra-small-xxxs"
         },
         {
           "id": "gid://shopify/TaxonomyValue/17115",
@@ -550,44 +540,14 @@
           "handle": "accessory-size__double-extra-small-xxs"
         },
         {
-          "id": "gid://shopify/TaxonomyValue/2916",
-          "name": "Extra large (XL)",
-          "handle": "accessory-size__extra-large-xl"
-        },
-        {
-          "id": "gid://shopify/TaxonomyValue/10616",
-          "name": "Extra long",
-          "handle": "accessory-size__extra-long"
-        },
-        {
-          "id": "gid://shopify/TaxonomyValue/10615",
-          "name": "Extra short",
-          "handle": "accessory-size__extra-short"
-        },
-        {
           "id": "gid://shopify/TaxonomyValue/2912",
           "name": "Extra small (XS)",
           "handle": "accessory-size__extra-small-xs"
         },
         {
-          "id": "gid://shopify/TaxonomyValue/17118",
-          "name": "Five extra large (5XL)",
-          "handle": "accessory-size__five-extra-large-5xl"
-        },
-        {
-          "id": "gid://shopify/TaxonomyValue/17117",
-          "name": "Four extra large (4XL)",
-          "handle": "accessory-size__four-extra-large-4xl"
-        },
-        {
-          "id": "gid://shopify/TaxonomyValue/2915",
-          "name": "Large (L)",
-          "handle": "accessory-size__large-l"
-        },
-        {
-          "id": "gid://shopify/TaxonomyValue/20089",
-          "name": "Long",
-          "handle": "accessory-size__long"
+          "id": "gid://shopify/TaxonomyValue/2913",
+          "name": "Small (S)",
+          "handle": "accessory-size__small-s"
         },
         {
           "id": "gid://shopify/TaxonomyValue/2914",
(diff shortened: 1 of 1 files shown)


test/models/attribute_test.rb
```diff
@@ -217,6 +217,57 @@ def teardown
     )
   end
 
+  test ".store_attribute_value_order_indexes stores ordered value indexes" do
```
elsom25, line 220: It feels a little off to me to be testing a private method? Either we just make this public or we test the right public interface that covers this 'side effect'?

KComrade53: I'm happy to remove it since its functionality is covered by the test right after it. I was just being thorough

KComrade53:
The amount of changes to the dist files for the other locales ended up being a few million lines of changes. So I will split that up into multiple PRs as to not overload GitHub

app/models/attribute.rb
```diff
@@ -83,8 +85,29 @@ def as_txt(attributes, version:, locale: "en")
...
     end
 
+    def sort_values(attribute_friendly_id, values, locale: "en")
+      if @custom_attribute_value_order_indexes&.dig(attribute_friendly_id)
+        return values.sort_by do |value|
+          @custom_attribute_value_order_indexes[attribute_friendly_id][value.friendly_id]
+        end
+      end
+
+      ValueSorter.sort_by_localized_name(values, locale: locale)
```
elsom25, line 95: I think the var extraction in the if ends up being easier to read? Unsure on the right var name for this one though 🤔 

```suggestion
      if (manual_sort = @custom_attribute_value_order_indexes&.dig(attribute_friendly_id))
        values.sort_by { manual_sort[_1.friendly_id] }
      else
        ValueSorter.sort_by_localized_name(values, locale: locale)
      end
```

app/models/attribute.rb
```diff
@@ -64,7 +64,9 @@ def find_localization(locale, id, key)
     #
     # `dist/` serialization
 
-    def as_json(attributes, version:, locale: "en")
+    def as_json(attributes, version:, locale: "en", raw_attributes:)
```
elsom25, line 67: `raw_attributes` isn't super intuitive to me in isolation. Is there a better representation within the context of the `Attribute` model?

app/models/regex_pattern.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+module RegexPattern
```
elsom25, line 3: This might be better inside an additional module? `RegexPattern` in isolation doesnt tell me much of what this actually is

app/models/value_sorter.rb
```diff
@@ -0,0 +1,82 @@
...
+      perform_sort(values, sort_by_proc)
+    end
+
+    private
+
+    def perform_sort(values, sort_by_proc)
+      sorted_values = values.sort do |a, b|
+        value_a = sort_by_proc.call(a)
+        value_b = sort_by_proc.call(b)
+
```
elsom25, line 17: The use of a proc here feels weird/unnecessary?

elsom25: Particularly b/c `perform_sort` is private and only used by this one method

app/models/regex_pattern.rb
elsom25: This class and the next one would both benefit with some heredocs for the classes themselves to help set context on purpose and use

app/models/value_sorter.rb
KComrade53: It's how it's handled in the gem. It allows it to be more flexible later if we want to use it elsewhere. But if we're okay with making it specific to this use case for now I can just update the sort method to access name directly and skip the extra step with the proc

app/models/attribute.rb
```diff
@@ -83,8 +85,31 @@ def as_txt(attributes, version:, locale: "en")
...
+
+      ordered_attribute_values = @custom_attribute_value_order_indexes[attribute_friendly_id]
+      values.sort_by do |value|
+        ordered_attribute_values[value.friendly_id]
+      end
+    end
+
     private
 
+    def store_custom_attribute_value_order_indexes(data_yml_base_attributes)
```
tylerrowsell, line 102: This feels like it is done at the wrong level. Everytime we want positional data we'll have to import custom information from data/attributes.yml

Could we make it so there is a `position` argument on values and when we seed the database set the position argument if it matches one of our special sort keys?

data/attributes.yml
```diff
@@ -13967,9 +13972,9 @@ base_attributes:
   values:
   - hardwood_lumber_grade__firsts_and_seconds_fas
   - hardwood_lumber_grade__no_1_common
-  - hardwood_lumber_grade__no_2_common
```
ricardotejedorsanz, line 13970: Edge case alpha is not handling well (2a before 2). Should we use custom here?

test/models/value_test.rb
```diff
@@ -66,14 +66,64 @@ def teardown
...
+        "primary_attribute_friendly_id" => "color",
+      },
     ]
 
+    base_attributes = [{ "id" => 1, "name" => "Color", "friendly_id" => "color" }]
+
     assert_difference -> { Value.count }, 2 do
-      Value.insert_all_from_data(data)
+      Value.insert_all_from_data(data, base_attributes)
+    end
```
tylerrowsell, line 91: Can we add an assertion that position is nil for these ones?

app/models/value.rb
```diff
@@ -69,13 +71,24 @@ def as_txt(values, version:, locale: "en")
 
     private
 
+    def row_from_data_with_position(data, base_attributes_by_friendly_id)
+      row = row_from_data(data)
+
+      matching_attribute = base_attributes_by_friendly_id[row["primary_attribute_friendly_id"]]
+
+      return row unless matching_attribute["sorting"] == "custom"
```
tylerrowsell, line 79: Nit: Does this need to be an enum? If we don't expect to add other sort types perhaps `matching_attribute["custom_sort"]` returns a boolean.

db/seed.rb
```diff
@@ -29,7 +29,7 @@ def import_taxonomy!(values_data:, attributes_data:, verticals_data:)
 
       vputs("  → Attributes and values")
       Attribute.insert_all_from_data(attributes_data["base_attributes"])
-      Value.insert_all_from_data(values_data)
+      Value.insert_all_from_data(values_data, attributes_data["base_attributes"])
```
tylerrowsell, line 32: I wonder if it would be clearer to write it like this.

```suggestion
    attributes_data["base_attributes"].each do |attribute|
      Value.insert_all_from_data(attribute.values, custom_sort: attribute[:custom_sorting])
    end
```

Then above 

```
  def insert_all_from_data(data, custom_sort)
     data.each_with_index { |value, index| value[:position] = index+1 } if custom_sort

      insert_all!(Array(data))
  end
```

tylerrowsell (approved):
Couple more suggestions to make it clearer but looks good.

db/seed.rb
KComrade53: Attribute values only stores the friendly_id of each value. The rest of the fields need to come from that values_data array

app/models/value.rb
KComrade53: I don't really want to prescribe how we're going to use it quite yet. If it evolves then this is an easy fix

data/attributes.yml
elsom25: Is this a tune to the regexes moreso? Sounds like we likely always want the generic form first?

PR #231, opened 2024-07-09
Add first batch of updated dist files with sorted attribute values

KComrade53:
Followup to https://github.com/Shopify/product-taxonomy/pull/230

Adds the first half of the updated dist json files after sorting each attribute's values with the logic added by that PR


tylerrowsell (commented):
~Why is there no `en.yml` updates?~ Nevermind

tylerrowsell (approved):

PR #236, opened 2024-07-11
Add latest to docs and true up our dist files

elsom25:


tylerrowsell (approved):

elsom25:
I cannot figure out what has changed and is making our automated checks have differing orders.. but will let that fix be it's own PR

PR #239, opened 2024-07-15
Localization context for linguists

tylerrowsell:
Closes https://github.com/Shopify/custom-data-issues/issues/2940

- Adds a script to generate english distributions from `/data` and add context for linguists.
- Updates make files to include sentinel for changes
- Updates english using above script.
- Updates all other localization files with comment that they're automatically generated via github action. (we're not there yet but will be soon).

### DIFF
--- a/dist/en/attributes.json
+++ b/dist/en/attributes.json
@@ -68228,14 +68228,14 @@
       ],
       "values": [
         {
-          "id": "gid://shopify/TaxonomyValue/16448",
-          "name": "0/1",
-          "handle": "national-regional-rating__0-1"
-        },
-        {
           "id": "gid://shopify/TaxonomyValue/22434",
           "name": "0",
           "handle": "national-regional-rating__0"
+        },
+        {
+          "id": "gid://shopify/TaxonomyValue/16448",
+          "name": "0/1",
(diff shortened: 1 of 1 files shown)


data/localizations/attributes/de.yml
```diff
@@ -1,3 +1,4 @@
+# This file is auto-generated through a github action. Do not edit directly.
```
KComrade53, line 1: Does this mean public contributors will no longer be able to submit translations suggestions via PR? Would they just do so via an issue then if they have translation suggestions?

tylerrowsell: Yes that's correct. We'll want to confirm any localization with the linguists so we'd prefer to go through an issue.

bin/sync_en_localizations
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require "yaml"
+require "fileutils"
+require_relative "../config/environment"
```
elsom25, line 6: We have a CLI helper — let's use that in this file as well?

elsom25 (approved):
Good to go, though would like to have the bin file use our CLI helper before we call it _done_

PR #242, opened 2024-07-18
Check missing mappings for Shopify categories

Catherine-Fu:
### What are you trying to accomplish? 
Checking which Shopify categories lack corresponding channel mappings

### Testing result
Screenshot 2024-07-22 at 6 29 25 PM

### DIFF
--- a/test/integration/mapping_validation_test.rb
+++ b/test/integration/mapping_validation_test.rb
@@ -3,9 +3,9 @@
 require_relative "../test_helper"
 
 class MappingValidationTest < ActiveSupport::TestCase
-  include Minitest::Hooks
   def setup
-    @mappings_json_data = CLI.new.parse_json("dist/en/integrations/all_mappings.json")
+    @sys = System.new
+    @mappings_json_data = @sys.parse_json("dist/en/integrations/all_mappings.json")
   end
 
   test "category IDs in mappings are valid" do
@@ -89,10 +89,8 @@ class MappingValidationTest < ActiveSupport::TestCase
   end
 
   def category_ids_from_taxonomy(input_or_output_taxonomy)
-    cli = CLI.new
-
     if input_or_output_taxonomy.include?("shopify") && !input_or_output_taxonomy.include?("shopify/2022-02")
-      categories_json_data = cli.parse_json("dist/en/categories.json")
+      categories_json_data = @sys.parse_json("dist/en/categories.json")
       shopify_category_ids = Set.new
       categories_json_data["verticals"].each do |vertical|
         vertical["categories"].each do |category|
@@ -103,7 +101,7 @@ class MappingValidationTest < ActiveSupport::TestCase
     else
       channel_category_ids = Set.new
       file_path = "data/integrations/#{input_or_output_taxonomy}/full_names.yml"
-      channel_taxonomy = sys.parse_yaml(file_path)
+      channel_taxonomy = @sys.parse_yaml(file_path)
       channel_taxonomy.each do |entry|
         channel_category_ids.add(entry["id"].to_s)
       end


test/integration/mapping_validation_test.rb
```diff
@@ -20,6 +20,35 @@ class MappingValidationTest < ActiveSupport::TestCase
       #{invalid_categories}"
   end
 
+  test "every Shopify category has corresponding channel mappings" do
+    mappings_json_data = CLI.new.parse_json("dist/en/integrations/all_mappings.json")
```
elsom25, line 24: Should we move this to a `before_all` to save parsing a very large file multiple times?

test/integration/mapping_validation_test.rb
```diff
@@ -20,6 +20,35 @@ class MappingValidationTest < ActiveSupport::TestCase
...
+    mappings_json_data["mappings"].each do |mapping|
+      next unless mapping["input_taxonomy"].include?("shopify")
+
+      all_shopify_category_ids = category_ids_from_taxonomy(mapping["input_taxonomy"])
+      next if all_shopify_category_ids.nil?
+
+      shopify_category_ids_from_mappings_input = Set.new
+      mapping["rules"].each do |rule|
+        shopify_category_ids_from_mappings_input.add(rule["input"]["category"]["id"])
+      end
```
elsom25, line 36: Can we do this as an operation instead of an each?
```suggestion
      shopify_category_ids_from_mappings_input = mapping["rules"]
        .map { _1.dig("input", "category", "id") }
        .to_set
```

test/integration/mapping_validation_test.rb
```diff
@@ -20,6 +20,35 @@ class MappingValidationTest < ActiveSupport::TestCase
...
+      shopify_category_ids_diff = all_shopify_category_ids - shopify_category_ids_from_mappings_input
```
elsom25, line 37: Should there be a return if this is empty?
```suggestion
      unmapped_category_ids = all_shopify_category_ids - shopify_category_ids_from_mappings_input
      next if unmapped_category_ids.empty?
```

test/integration/mapping_validation_test.rb
```diff
@@ -20,6 +20,35 @@ class MappingValidationTest < ActiveSupport::TestCase
...
+        input_taxonomy: mapping["input_taxonomy"],
+        output_taxonomy: mapping["output_taxonomy"],
+        count_of_shopify_category_ids_lack_mappings: shopify_category_ids_diff.size,
+        shopify_category_ids_lack_mappings: shopify_category_ids_diff,
+      }
+    end
+
+    assert shopify_category_lack_mappings.empty?,
+      "The following shopify category ids lack corresponding channel mappings:
+    #{shopify_category_lack_mappings}"
```
elsom25, line 49:
```suggestion
    assert_empty(
      shopify_category_lack_mappings,
      "The following shopify category ids lack corresponding channel mappings: #{shopify_category_lack_mappings}"
    )
```

elsom25: I also wonder if we should do this _in the loop_ vs collecting all of them? Or maybe fail on first to save time when there is failure? 🤔

elsom25 (commented):
Some code style thoughts

test/integration/mapping_validation_test.rb
Catherine-Fu: I addressed all the other comments.  `before_all` seems not applicable to this test suite because it aims to speed up testing for specs that share expensive database setup code according to the [minitest-hooks](https://github.com/jeremyevans/minitest-hooks) but this test suite does not involve db setup.  Instead, I defined `def setup` to include steps shared by different test cases.

test/integration/mapping_validation_test.rb
Catherine-Fu: Users need to get a full list of missing mappings to correct the issue throughly and they can run this test locally to avoid unnecessary CI checks.

test/integration/mapping_validation_test.rb
elsom25: I think parsing a massive YAML file _is slow_. Happy to be wrong on it speeding things up, but we use it for _exactly this_ in our other integration test

elsom25:
The test seems to be working as expected — though again we cannot merge this until we have the found conflicts resolved (and likely have tooling to keep these up to do date easily enough) (cc @jeanrick per our chats)

test/integration/mapping_validation_test.rb
```diff
@@ -19,6 +23,42 @@ class MappingValidationTest < ActiveSupport::TestCase
...
+    unless shopify_categories_lack_mappings.empty?
+      puts "Shopify Categories are missing mappings for the following integrations:"
+      shopify_categories_lack_mappings.each_with_index do |mapping, index|
+        puts ""
+        puts "[#{index + 1}] #{mapping[:input_taxonomy]} to #{mapping[:output_taxonomy]} (#{mapping[:unmapped_category_ids].size} missing)"
+        mapping[:unmapped_category_ids].each do |category_id|
+          puts " - #{category_id}"
+        end
+      end
+      assert(shopify_categories_lack_mappings.empty?, "Shopify Categories are missing mappings.")
```
elsom25, line 58: nit:
```suggestion
      assert_predicate(shopify_categories_lack_mappings, :empty?, "Shopify Categories are missing mappings.")
```

Catherine-Fu:
This PR was merged automatically as part of this PR: [Excluded given taxonomy for mappings in checking missing mappings](https://github.com/Shopify/product-taxonomy/pull/250)

PR #244, opened 2024-07-22
Category generator script

elsom25:
- Extracted Alphanumeric sorting to lib (also made sorting stable)
- Adds bin/data commands to generate new categories and rebuild all data files from DB
- Updated Category + ValueSorter accordingly

```bash
> bin/data/add_category --name "Sweet Sweaters" --parent_id="aa"
> bin/data/dump_categories
```

### DIFF
--- a/bin/data/add_category
+++ b/bin/data/add_category
@@ -1,0 +1,35 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require_relative "../../config/environment"
+require_relative "../../config/cli"
+
+@cli = CLI.new([:name, :id, :parent_id]) do |opts|
+  opts.on("-n NAME", "--name=NAME", String, "Name of new category")
+  opts.on("-p PARENT_ID", "--parent_id=PARENT_ID", String, "Parent ID")
+  opts.on("-i ID", "--id=ID", String, "Statically defined ID")
+end
+@cli.parse!(ARGV)
+@cli.options_status
+
+fail "Missing required option: --name" unless @cli.options.name
+fail "Missing required option: --parent_id" unless @cli.options.parent_id
+
+parent = Category.find_by(id: @cli.options.parent_id)
+fail "Parent category <#{@cli.options.parent_id}> not found" unless parent
+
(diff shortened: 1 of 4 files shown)


bin/data/add_category
KComrade53, line 1: I wonder if it would be more valuable to make an interactive CLI script that we can utilize for all kinds of adds and changes. So they just have to follow along the prompts instead of passing args and repeatedly calling commands

KComrade53: This is a good start though and I'm happy for us to start with something like this in the meantime. It's a lot better than nothing

lib/alphanumeric_sorter.rb
```diff
@@ -0,0 +1,113 @@
...
+          *normalize_value(value),
+          idx
+        ]
+      end
+    end
+
+    def normalize_value(value)
+      @normalized_values ||= {}
+      @normalized_values[value] ||= begin
+        if (numerical = value.match(RegexPattern::NUMERIC_PATTERN))
```
KComrade53, line 18: This is an anti-pattern and I declared variables above instead to avoid it. Also ends up a bit more readable. I'm surprised RuboCop isn't complaining about this, because it usually does

bin/data/dump_categories
KComrade53, line 1: What's the use case for this script?

KComrade53 (commented):
A step in the right direction to making contributions easier 😄

lib/alphanumeric_sorter.rb
elsom25: I'd say it is more about the complexity of the call. Rubocop specifically complains when you don't have the `()` (which is the soft signal assignment is happening).

I opted for this pattern here to save a regex check :grin:

KComrade53 (approved):

lib/alphanumeric_sorter.rb
KComrade53: If you want to keep it we can keep it. GPT suggested the same thing to me when I was writing the initial sorter. 

Anti-pattern aside I liked the readability of `if numerical do, elsif sequential do, else`

PR #246, opened 2024-07-23
Add bin/data/rename_category

elsom25:
Renames categories, including en translations. Able to rename root categories as well, which impacts files.

**Renames** should not require mappings work. **Delete** (and **Changes** aka delete + add) will.

Example renaming root node back (had previously renamed to `Appartments`):
```bash
# bin/data/rename_category --id="aa" --name="Apparel & Accessories" --verbose

Options: {:name=>"Apparel & Accessories", :id=>"aa", :verbose=>true, :force=>false}
Updated category `aa` to `Apparel & Accessories` from `Appartments` in DB
Renaming category file from `aa_appartments.yml` to `aa_apparel_accessories.yml`
→ Moving `data/categories/aa_appartments.yml` to `data/categories/aa_apparel_accessories.yml`
→ Writing `data/categories/aa_apparel_accessories.yml`
→ Running `bin/sync_en_localizations`
```

### DIFF
--- a/app/models/category.rb
+++ b/app/models/category.rb
@@ -153,8 +153,9 @@ class Category < ApplicationRecord
     ancestors.reverse.map { _1.name(locale:) }.push(name(locale:)).join(" > ")
   end
 
+  # should never use translations
   def handleized_name
-    "#{id}_#{name.downcase.gsub(%r{[^a-z0-9\s\-_/\.\+#]}, "").gsub(/[\s\-\.]+/, "_")}"
+    "#{id}_#{self[:name].downcase.gsub(%r{[^a-z0-9\s\-_/\.\+#]}, "").gsub(/[\s\-\.]+/, "_")}"
   end
 
   def root?
@@ -210,7 +211,7 @@ class Category < ApplicationRecord
   def as_json_for_data
     {
       "id" => id,
-      "name" => name,
+      "name" => self[:name], # avoid localization
       "children" => children.sort_by(&:id_parts).map(&:id),
       "attributes" => AlphanumericSorter.sort(related_attributes.map(&:friendly_id), other_last: true),
     }
--- a/bin/data/rename_category
+++ b/bin/data/rename_category
@@ -1,0 +1,37 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require_relative "../../config/environment"
+require_relative "../../config/cli"
+
+@cli = CLI.new([:name, :id]) do |opts|
+  opts.on("-n NAME", "--name=NAME", String, "New name for the category")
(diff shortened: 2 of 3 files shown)


bin/data/rename_category
```diff
@@ -0,0 +1,37 @@
...
+
+original_name = category[:name] # skip localizations
+original_handle = category.handleized_name
+category.update!(name: @cli.options.name)
+@cli.vputs("Updated category `#{category.id}` to `#{category[:name]}` from `#{original_name}` in DB")
+
+if category.handleized_name != original_handle
+  @cli.vputs("Renaming category file from `#{original_handle}.yml` to `#{category.handleized_name}.yml`")
+  @cli.move_file!("data/categories/#{original_handle}.yml", "data/categories/#{category.handleized_name}.yml")
+end
```
KComrade53, line 28: Why does updating a category's name cause files to be moved? That would only be necessary if we're renaming a root category and therefore vertical. Will this not error for renaming non-root categories?

I assume most, if not all, renames will be on non-root categories

app/models/category.rb
```diff
@@ -210,7 +211,7 @@ def related_attribute_friendly_ids=(ids)
   def as_json_for_data
     {
       "id" => id,
-      "name" => name,
+      "name" => self[:name], # avoid localization
```
KComrade53, line 214: I assume this means there's wip for changing localizations or something? Unless I'm mistaken, all the scripts using this only work for the english locale anyways

bin/data/rename_category
```diff
@@ -0,0 +1,37 @@
...
+@cli.parse!(ARGV)
+@cli.options_status
+
+fail "Missing required option: --name" unless @cli.options.name
+fail "Missing required option: --id" unless @cli.options.id
+
+category = Category.find_by(id: @cli.options.id)
+fail "Category `#{@cli.options.id}` not found" unless category
+
+original_name = category[:name] # skip localizations
```
KComrade53, line 20: Seems redundant. Looks to me like this script only works on English anwyays

bin/data/rename_category
elsom25: I mean, we can fail on verticals too, but notionally a strict rename is allowed at that layer (this isn't a move, its really meant for spelling mistakes)

app/models/category.rb
elsom25: Nothing WIP on my side; just finding spots where we need to turn it off — keeping that targeted for the moment until we see more (but adding a comment so folks understand why I'm not using the method, since it'll be surprising to a casual reader)

bin/data/rename_category
elsom25: This is more for active dev than _running as normal_. Seeing the raw value is helpful here IMO

PR #248, opened 2024-07-24
Use CLI::UI for bin/

elsom25:
Use `CLI::UI` for output, and `TTY::Options` for richer option parsing.

Ported all bin commands into `app/commands/*`


app/commands/application_command.rb
```diff
@@ -0,0 +1,112 @@
...
+  end
+
+  class << self
+    def run(argv = ARGV)
+      new(argv:, interactive: true).run
+    end
+  end
+
+  def initialize(argv: ARGV, interactive: false, **kwargs)
+    super()
```
KComrade53, line 37: What `super` is this invoking? This class doesn't extend another class

app/commands/rename_data_category_command.rb
```diff
@@ -0,0 +1,68 @@
...
+    validate { _1 =~ Category::ID_REGEX }
+  end
+
+  keyword :name do
+    desc "Updated category name"
+    required
+  end
+
+  def execute
+    frame("Adding new category") do
```
KComrade53, line 20: This should say "Renaming category" or the like

app/commands/rename_data_category_command.rb
```diff
@@ -0,0 +1,68 @@
...
+      @category.update!(name: params[:name])
+
+      sp.update_title("Updated category `#{original_name}` to `#{params[:name]}`")
+    end
+  end
+
+  def update_vertical_file!
+    spinner("Updating vertical file") do |sp|
+      if @category.root?
+        sys.move_file!("data/categories/#{@original_handle}.yml", "data/categories/#{@category.handleized_name}.yml")
```
KComrade53, line 51: I think this should be logged when it happens so the user is informed that the file is being renamed

app/commands/new_data_category_command.rb
```diff
@@ -0,0 +1,73 @@
...
+
+class NewDataCategoryCommand < ApplicationCommand
+  usage do
+    no_command
+  end
+
+  keyword :name do
+    desc "Category name to create"
+    required
+  end
```
KComrade53, line 11: Looks like these change how data is passed to the command. Want to add an example usage to your PR description so this is highlighted?

Old method: `--name="foo"`, new method: `NAME="foo"`

elsom25: Yes intentional — haven't documented any of these commands yet, and figure they'll change again as we net out a few more of these

elsom25: I'll make sure we have proper docs for these though — after which changes will require update to those 😁

PR #255, opened 2024-07-31
Update att desc

tylerrowsell:
Closes https://github.com/Shopify/custom-data-issues/issues/2851

The last update of the missing attribute descriptions and their localizations.

@elsom25 just to make sure I got this right... everything on `main` dists should be unstable right? I think this was a mistake I made when releasing `2024-07`

### DIFF
--- a/dist/en/attribute_values.json
+++ b/dist/en/attribute_values.json
@@ -1,5 +1,5 @@
 {
-  "version": "2024-07",
+  "version": "2024-10-unstable",
   "values": [
     {
       "id": "gid://shopify/TaxonomyValue/13254",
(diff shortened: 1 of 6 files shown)


missing_attributes.csv
KComrade53, line 1: Did you mean to commit this?

script.rb
KComrade53, line 1: Did you mean to commit this script?

elsom25:
>everything on main dists should be unstable right?

Yep generally. We "tag" the moment that a release exists and main continues forward

VERSION
```diff
@@ -1 +1 @@
-2024-07
+unstable
```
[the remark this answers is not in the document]
elsom25: Or maybe `2024-07-unstable`? Or we keep it as raw `unstable`

PR #259, opened 2024-08-02
[WIP] Sharing a start of some cleanup

elsom25:
- Lean more heavily on TTY::Option for all args
- Make more of the client config, configurable
- Start reordering methods to naturally follow reading order
- Leverage ActiveRecord a touch more
- Update spinner titles as things happen

Only made it through a few shifts, but hopefully helps give a bit better signal 😁

### DIFF
--- a/app/commands/generate_missing_mappings_command.rb
+++ b/app/commands/generate_missing_mappings_command.rb
@@ -4,6 +4,56 @@ class GenerateMissingMappingsCommand < ApplicationCommand
   UNSTABLE = "unstable"
   EMBEDDING_MODEL = "text-embedding-3-small"
   MAPPING_GRADER_GPT_MODEL = "gpt-4"
+  SYSTEM_PROMPT = <<~PROMPT
+    You are a taxonomy mapping expert who evaluate the accuracy of product category mappings between two taxonomies.
+    Your task is to review and grade the accuracy of the mappings, Yes or No, based on the following criteria:
+    1. Mark a mapping as Yes, i.e. correct, if two categories of a mapping are highly relevant to each other and similar
+      in terms of product type, function, or purpose.
+      For example:
+        - "Apparel & Accessories" and "Clothing, Shoes & Jewelry"
+        - "Apparel & Accessories > Clothing > One-Pieces" and "Clothing, Shoes & Accessories > Women > Women's Clothing > Jumpsuits & Rompers"
+    2. Mark a mapping as No, i.e. incorrect, if two categories of a mapping are irrevant to each other
+      in terms of product type, function, or purpose.
+      For example:
+        - "Apparel & Accessories > Clothing > Dresses" and "Clothing, Shoes & Jewelry>Shoe, Jewelry & Watch Accessories"
+        - "Apparel & Accessories" and "Clothing, Shoes & Jewelry>Luggage & Travel Gear"
+    Note, the character ">" in a category name indicates the start of a new category level. For example:
+    "sporting goods > exercise & fitness > cardio equipment"'s ancestor categories are "sporting goods > exercise & fitness" and "sporting goods".
+    You will receive a list of mappings. Each mapping contains a from_category name and a to_category name.
+    e.g. user's prompt in json format:
+    [
+      {
+        "from_category_id": "111",
+        "from_category": "Apparel & Accessories > Jewelry > Smart Watches",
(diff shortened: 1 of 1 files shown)


app/commands/generate_missing_mappings_command.rb
```diff
@@ -1,68 +1,116 @@
...
+  option :retries do
+    desc "Number of retries for OpenAI API"
+    short "-r"
+    long "--retries integer"
+    default 3
+    convert :to_i
+  end
+
   def execute
+    @embeddings = {}
```
elsom25, line 43: Was a thought on if this could save passing certain variables around all the time... but its unfinished so might not be a keeper

app/commands/generate_missing_mappings_command.rb
```diff
@@ -1,68 +1,116 @@
...
-          [id, category_full_name] if category_full_name
-        end.compact.to_h
+    spinner("Finding Shopify categories that are unmapped") do
+      all_shopify_ids = Set.new(Category.all.pluck(:id))
+      mappings_by_output = MappingRule
+        .where(input_version: "shopify/#{params[:shopify_version]}")
+        .group_by(&:output_version)
+      @unmapped_categories = mappings_by_output.filter_map do |output_taxonomy, mappings|
+        # Can this be an ActiveRecord call? Should it?
+        # TODO: Where is #product_category_id defined? AFAICT #input returns a hash
```
elsom25, line 72: `#input` returns a `product` maybe? But that also doesn't have this method 🤔

app/commands/generate_missing_mappings_command.rb
```diff
@@ -1,68 +1,116 @@
...
+        found_shopify_ids = Set.new(mappings.map { _1.input.product_category_id.split("/").last })
+        unmapped_ids = all_shopify_ids - found_shopify_ids
+        category_ids_full_names = Category.where(id: unmapped_ids).pluck(:id, :full_name).to_h
         next if category_ids_full_names.empty?
 
-        @unmapped_category_groups << {
-          input_taxonomy: mappings.first.input_version,
```
elsom25, line 45: This is constant, so feels okay to leave off? (given the added heading)

app/commands/generate_missing_mappings_command.rb
```diff
@@ -1,68 +1,116 @@
...
+          )
+        end
       end
     end
   end
 
   def load_embedding_data(output_taxonomy)
-    embeddings = nil
-    spinner("Loading embeddings for #{output_taxonomy}") do
+    spinner("Loading embeddings for #{output_taxonomy}") do |sp|
```
elsom25, line 102: spinners return the results of their block; lets us save the var passing

elsom25: Also given the frame, we can probably drop `for #{output_taxonomy}` for everything called within said frame

app/commands/generate_missing_mappings_command.rb
chesterbot01: I thought about making embeddings an instance variable, but they are specific to each channel integration (though, at the moment, there's only one: Google)

app/commands/generate_missing_mappings_command.rb
chesterbot01: > TODO: Where is #product_category_id defined?

It was defined in shopify_product.rb as `store :payload, coder: JSON, accessors: [:product_category_id, :properties]`

app/commands/generate_missing_mappings_command.rb
```diff
@@ -1,185 +1,236 @@
...
-
-        @unmapped_category_groups << {
-          input_taxonomy: mappings.first.input_version,
-          output_taxonomy: output_version,
-          category_ids_full_names: category_ids_full_names,
-        }
+    spinner("Finding Shopify categories that are unmapped") do
+      all_shopify_ids = Set.new(Category.all.pluck(:id))
+      mappings_by_output = MappingRule
+        .where(input_version: "shopify/#{params[:shopify_version]}")
```
chesterbot01, line 66: Is it possible to move version-related changes out of the tooling scope? This will cause the tool to break.

elsom25: For now though, have remove the file parsing, which will ensure this stays as just `unstable` until we're ready for broader changes

PR #267, opened 2024-08-07
Add a workflow to trigger taxonomy mapping generation

chesterbot01:
Added a command `/generate_mappings` to trigger the generation of taxonomy mappings.
See the linked PR for details of how it works.

### DIFF
--- a/.github/workflows/generate_missing_mappings.yml
+++ b/.github/workflows/generate_missing_mappings.yml
@@ -11,7 +11,7 @@ concurrency:
 jobs:
   generate_missing_mappings:
     if: |
-      github.repository_owner == 'Shopify' &&
+      contains('["OWNER", "COLLABORATOR", "MEMBER"]', github.event.comment.author_association) &&
       github.event.issue.pull_request &&
       github.event.comment.body == '/generate_mappings'
     runs-on: ubuntu-latest
--- a/.github/workflows/generate_dist.yml
+++ b/.github/workflows/generate_dist.yml
@@ -12,7 +12,7 @@ concurrency:
 jobs:
   generate_dist:
     if: |
-      github.repository_owner == 'Shopify' &&
+      contains('["OWNER", "COLLABORATOR", "MEMBER"]', github.event.comment.author_association) &&
       github.event.issue.pull_request &&
       github.event.comment.body == '/generate_dist'
     runs-on: ubuntu-latest


elsom25:
@chesterbot01 the pattern in the generate_dist command that's been re-added should work for us here as well. Key is segregating `write` token access

.github/workflows/generate_missing_mappings.yml
```diff
@@ -0,0 +1,113 @@
...
+    - name: Generate missing mappings
+      env:
+        OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+        OPENAI_API_BASE: https://openai-proxy.shopify.ai/v1
+      run: |
+        make generate_mappings
+    - name: Clean cache and re-generate mapping distribution files
+      run: |
+        make clean
+        make VERBOSE=1
```
elsom25, line 53: Curious why we need a make build before and after the generate mappings call?

.github/workflows/generate_missing_mappings.yml
```diff
@@ -0,0 +1,113 @@
...
+        echo "Waiting for Qdrant server to be ready..."
+        sleep 25
+    - uses: cue-lang/setup-cue@v1.0.0
+      with:
+        version: 'v0.7.0'
+    - name: Generate distribution files to gather all published mappings in json format
+      run: make VERBOSE=1
+    - name: Generate missing mappings
+      env:
+        OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
```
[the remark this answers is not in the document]
elsom25: Basically anything we expose to env and to `make` is accessible. Maybe we force this to to run against the `Makefile` in main? That way it's a known quantity.

.github/workflows/generate_missing_mappings.yml
elsom25: Maybe better to use `bin/generate_dist` directly with a targets flag for the second call? And the initial one could be `make seed` to save running generation logic when we don't need it?

chesterbot01: > use bin/generate_dist directly with a targets flag for the second call?

The `make generate_mappings` command in the initial stage modifies yml files in the `data/integrations` directory. However, the follow-up `bin/generate_dist` command doesn't seed these recent changes because it's using a cached version of the database.

> the initial one could be `make seed`

Sounds good.

chesterbot01: Done.

.github/workflows/generate_missing_mappings.yml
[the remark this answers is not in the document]
elsom25: >Or we can let the workflow return early if detects that Makefile changes in the PR
I like that actually (not sure what you picked). Feels simpler and easy to explain. We should mirror this to our other generate command as well

.github/workflows/generate_missing_mappings.yml
```diff
@@ -0,0 +1,118 @@
...
+    - uses: cue-lang/setup-cue@v1.0.0
+      with:
+        version: 'v0.7.0'
+    - name: Seed database
+      run: make seed
+    - name: Override Makefile with main branch version
+      run: |
+        git fetch origin main
+        git checkout origin/main -- Makefile
+        echo "Makefile has been overridden with the version from the main branch"
```
elsom25, line 48: Should we move this to the top? Basically enforce all make commands only run through main's version? Also lets us clump it next to the other git command which feels good for readability

elsom25: And when making that change, want to port it to the generate dist too?

.github/workflows/generate_missing_mappings.yml
```diff
@@ -0,0 +1,118 @@
...
+        echo "$commit_output"
+        echo "::set-output name=commit_message::$commit_output"
+        git push || echo "No changes to push"
+    - name: Read disagree mapping entries
+      id: read_content
+      run: |
+        if [ -f tmp/mappings_with_low_confidence.txt ]; then
+          low_confidence_mappings=$(cat tmp/mappings_with_low_confidence.txt)
+        else
+          low_confidence_mappings="No meaningful content provided."
```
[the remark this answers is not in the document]
chesterbot01: Right, fixed.

.github/workflows/generate_missing_mappings.yml
```diff
@@ -0,0 +1,163 @@
...
+        curl -L https://github.com/BLAKE3-team/BLAKE3/releases/download/1.5.3/b3sum_linux_x64_bin -o b3sum
+        chmod +x b3sum
+        sudo mv b3sum /usr/local/bin/
+
+    - name: Seed database
+      run: make --file=Makefile seed
+    - name: Generate missing mappings
+      env:
+        OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+        OPENAI_API_BASE: https://openai-proxy.shopify.ai/v1
```
elsom25, line 63: Let's change this to be a GH env-var

.github/workflows/generate_missing_mappings.yml
```diff
@@ -0,0 +1,163 @@
...
+        curl -L https://github.com/BLAKE3-team/BLAKE3/releases/download/1.5.3/b3sum_linux_x64_bin -o b3sum
+        chmod +x b3sum
+        sudo mv b3sum /usr/local/bin/
+
+    - name: Seed database
+      run: make --file=Makefile seed
+    - name: Generate missing mappings
+      env:
+        OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+        OPENAI_API_BASE: ${{ secrets.OPENAI_API_KEY }}
```
elsom25, line 63:
```suggestion
        OPENAI_API_BASE: ${{ secrets.OPENAI_API_BASE }}
```

.github/workflows/generate_missing_mappings.yml
```diff
@@ -11,7 +11,7 @@ concurrency:
 jobs:
   generate_missing_mappings:
     if: |
-      github.repository_owner == 'Shopify' &&
+      contains('["OWNER", "COLLABORATOR", "MEMBER"]', github.event.comment.author_association) &&
```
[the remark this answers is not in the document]
chesterbot01: Done.

PR #269, opened 2024-08-07
Adjusting mapping versions in the /dist folder

Catherine-Fu:
- [ ] ~~Mapping files under the the dist folder should be at the version we are currently targeting without the `-unstable` because the mappings under the dist folder is a running release~~.
- [x] Update mappings versions in the dist folder to reflect what @main branch represents.
- [x] Only keep the latest versioned mappings in dist folder and automatically delete old versioned mappings in dist folder when we run `make build`
- [x] Check that Shopify taxonomy version is in consistent between VERSION file and mappings in the /data folder

### DIFF
--- a/app/commands/generate_dist_command.rb
+++ b/app/commands/generate_dist_command.rb
@@ -112,7 +112,7 @@ class GenerateDistCommand < ApplicationCommand
 
   def generate_mapping_group_files(locale, records)
     directory_path = "dist/#{locale}/integrations/#{records.first.integration.name}"
-    sys.delete_files!(directory_path) if directory_path.present?
+    sys.delete_files!(directory_path)
 
     input_version = records.first.input_version.gsub("/", "_")
     output_version = records.first.output_version.gsub("/", "_")
--- a/lib/system.rb
+++ b/lib/system.rb
@@ -86,7 +86,7 @@ class System
 
   def delete_files!(directory_path_from_root)
     logger.debug("→ Deleting files under directory `#{directory_path_from_root}`")
-    FileUtils.rm_rf(Dir.glob("#{directory_path_from_root}/*"))
+    FileUtils.rm_rf(Dir.glob(path_for("#{directory_path_from_root}/*")))
     logger.success("Deleted files under directory`#{directory_path_from_root}`")
   end
 


elsom25:
>Mapping files under the the dist folder should be at the version we are currently targeting without the -unstable because the mappings under the dist folder is a running release.

I think some broken telephone is emerging 😅 

`main` _is unstable_, and that should apply to dist files too. The only stable moments are when we're cutting a release

app/commands/seed_local_command.rb
```diff
@@ -13,12 +13,6 @@ class SeedLocalCommand < ApplicationCommand
     permit ["taxonomy", "integrations"]
   end
 
-  option :version do
-    desc "Distribution version"
-    short "-V"
-    long "--version string"
-  end
```
elsom25, line 20: This does feel like a good change irregardless of my root comment though 😁

dist/en/integrations/google/shopify_2024-10_to_google_2021-09-21.json
```diff
@@ -1,8 +1,8 @@
 {
-  "version": "2024-10-unstable",
+  "version": "2024-10",
```
elsom25, line 2: This should be `2024-10-unstable`

dist/en/integrations/shopify/shopify_2022-02_to_shopify_2024-10.txt
```diff
@@ -1,4 +1,4 @@
-# Shopify Product Taxonomy - Mapping shopify/2022-02 to shopify/2024-07
+# Shopify Product Taxonomy - Mapping shopify/2022-02 to shopify/2024-10
```
elsom25, line 1: Now these ones though — should this not be `2024-07 to 2024-10-unstable`? cc @jeanrick

jeanrick: Yea, if we're including the `unstable` labels in `dist` then this version should be `2024-10-unstable`. The file name will also change from `shopify_2022-02_to_shopify_2024-07.txt` to `shopify_2022-02_to_shopify_2024-10-unstable.txt`

Catherine-Fu:
Hi @elsom25, I updated the mappings versions in the dist folder to reflect what @main branch represents.

dist/en/integrations/shopify/shopify_2022-02_to_shopify_2024-10.txt
Catherine-Fu: An alternative would be reserve `unstable` only in file contents but removing the `unstable` from file names.   Then when we cut a new release e.g., from `2024-10-unstable` to `2024-10`,  the file names would be unchanged instead of replacing by new files where the diff would include the whole file contents.   If the file names are unchanged, all diff in the release PR are about version value changes in files.  

If that makes sense to you, I'll implement it. cc @jeanrick @elsom25

Catherine-Fu: we can add the following code after https://github.com/Shopify/product-taxonomy/blob/e9d8dbe5cd0f6c6eb7914bb0482ba9e9acc781a3/app/commands/generate_dist_command.rb#L115-L116
```
    if input_version.include?("-unstable")
      input_version = input_version.delete_suffix("-unstable")
    end

    if output_version.include?("-unstable")
      output_version = output_version.delete_suffix("-unstable")
    end
```

Then https://github.com/Shopify/product-taxonomy/blob/e9d8dbe5cd0f6c6eb7914bb0482ba9e9acc781a3/app/commands/generate_dist_command.rb#L119-L120  will not include "unstable" to mapping file names

app/commands/generate_dist_command.rb
```diff
@@ -112,6 +112,8 @@ def generate_mapping_files(locale)
 
   def generate_mapping_group_files(locale, records)
     directory_path = "dist/#{locale}/integrations/#{records.first.integration.name}"
+    FileUtils.rm_rf(Dir.glob("#{directory_path}/*")) if directory_path.present?
```
elsom25, line 115: Should this be part of `make clean` or is it right for this to be part of _dist_?

If we keep this, can we find a way to shift parts of this into `System`. Trying to keep all file ops wrapped by that class

jeanrick: +1: this is a good utility to have in `System`

test/integration/mapping_validation_test.rb
```diff
@@ -111,6 +111,37 @@ def setup
     end
   end
 
+  test "Shopify taxonomy version is in consistent between VERSION file and mappings in the /data folder" do
+    shopify_taxonomy_version_from_file = "shopify/" + @sys.read_file("VERSION").strip
```
jeanrick, line 115: _nit_ Not a blocker for the PR, but it would be nice to have a more standard way of getting the "currently active version"

test/integration/mapping_validation_test.rb
```diff
@@ -111,6 +111,37 @@ def setup
...
+    mapping_rule_files = @sys.glob("data/integrations/*/*/mappings/*_shopify.yml")
+    files_include_inconsistent_shopify_taxonomy_version = []
+    mapping_rule_files.each do |file|
+      raw_mappings = @sys.parse_yaml(file)
+      [raw_mappings["input_taxonomy"], raw_mappings["output_taxonomy"]].each do |taxonomy_version|
+        next if !taxonomy_version.include?("shopify") || taxonomy_version.include?("shopify/2022-02")
```
jeanrick, line 121: Similar to above comment, ideally we shouldn't be doing YAML file loading/parsing here to check the taxonomy version and instead should be building these kinds of checks against our models

jeanrick (approved):
Changes look good overall
> Left a few comments on some general improvements I think we need to make, but not a blocker for this PR

dist/en/integrations/shopify/shopify_2022-02_to_shopify_2024-10.txt
jeanrick: I can see the merit of clearly saying "these files are unstable" via the file name, but I'm not a fan of the whole file getting replaced when we cut the release 🤔 

But given that we don't do this for the other taxonomy files, I lean to keeping it out of the filename for now

app/commands/generate_dist_command.rb
Catherine-Fu: Done

lib/system.rb
```diff
@@ -84,6 +84,12 @@ def delete_file!(path_from_root)
     logger.success("Deleted `#{path_from_root}`")
   end
 
+  def delete_files!(directory_path_from_root)
+    logger.debug("→ Deleting files under directory `#{directory_path_from_root}`")
+    FileUtils.rm_rf(Dir.glob("#{directory_path_from_root}/*"))
```
elsom25, line 89: Please use `path_for` for everything in this class (ensures we're scoped properly).

test/integration/mapping_validation_test.rb
Catherine-Fu: > checks against our models 

As we are doing checks in tests and the DB states are empty when accessing models e.g., MappingRule from tests, that's why I load data from YAML files directly.

lib/system.rb
Catherine-Fu: Done

app/commands/generate_dist_command.rb
```diff
@@ -112,7 +112,7 @@ def generate_mapping_files(locale)
 
   def generate_mapping_group_files(locale, records)
     directory_path = "dist/#{locale}/integrations/#{records.first.integration.name}"
-    FileUtils.rm_rf(Dir.glob("#{directory_path}/*")) if directory_path.present?
+    sys.delete_files!(directory_path) if directory_path.present?
```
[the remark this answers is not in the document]
Catherine-Fu: removed `if directory_path.present?`

PR #274, opened 2024-08-13
[AA] Add toddler sizes to Size attribute

KComrade53:
Adds age-range toddler sizes to the Size attribute


tylerrowsell (commented):
Changes make sense... we are figuring out the translations piece going forward so may need to hold until that lands.

ricardotejedorsanz (approved):
I agree Size is the right attribute for these changes. I suggest we add a few more missing values in the Baby/Toddler size range.

tylerrowsell (approved):

ricardotejedorsanz:
Moving to https://github.com/Shopify/product-taxonomy/pull/390 to override CLA issue.

PR #275, opened 2024-08-13
Add command for easily adding values

KComrade53:
Add a new command for easily adding values to the yml files.

Example usage:
`bin/data/add_value attribute_friendly_id="size" name="Big"`.

This command was used to generate the changes for https://github.com/Shopify/product-taxonomy/pull/274

### DIFF
--- a/app/commands/source/add_value_command.rb
+++ b/app/commands/source/add_value_command.rb
@@ -18,15 +18,8 @@ module Source
 
     def execute
       frame("Adding new value") do
-        ActiveRecord::Base.transaction do
-          setup!
-          create_value!
-          create_attributes_value!
-        rescue ActiveRecord::Rollback => e
-          logger.fatal(e.message)
-          exit(1)
-        end
-
+        setup!
+        create_value!
         update_data_files!
       end
     end
@@ -36,34 +29,19 @@ module Source
     def setup!
       @primary_attribute = Attribute.find_by(friendly_id: params[:attribute_friendly_id])
 
-      raise ActiveRecord::Rollback,
-        "Primary attribute `#{params[:attribute_friendly_id]}` not found" if @primary_attribute.nil?
+      if @primary_attribute.nil?
+        logger.fatal("Primary attribute `#{params[:attribute_friendly_id]}` not found")
+        exit(1)
+      end
 
       @extended_attributes = @primary_attribute.extended_attributes
     end
 
     def create_value!
-      @new_value = Value.find_or_create_for_attribute(@primary_attribute, params[:name])
-
-      return if @new_value.persisted?
-
-      raise ActiveRecord::Rollback, "Failed to create value: #{@new_value.errors.full_messages.to_sentence}"
-    end
-
-    def create_attributes_value!
-      attributes = [@primary_attribute, *@extended_attributes]
-
-      attributes.each do |attribute|
-        attributes_value = AttributesValue.find_or_create_by(
-          related_attribute: attribute,
-          value: @new_value,
-        )
-
-        next if attributes_value.persisted?
-
-        raise ActiveRecord::Rollback,
-          "Failed to link value to attribute: #{attributes_value.errors.full_messages.to_sentence}"
-      end
+      Value.find_or_create_for_attribute!(@primary_attribute, params[:name])
+    rescue => e
+      logger.fatal("Failed to create value: #{e.message}")
+      exit(1)
     end
 
     def update_data_files!
--- a/app/models/attribute.rb
+++ b/app/models/attribute.rb
@@ -148,7 +148,7 @@ class Attribute < ApplicationRecord
     sorting == "custom"
   end
 
-  def next_position
+  def next_value_position
     return unless manually_sorted?
 
     values.max_by(&:position).position + 1
--- a/app/models/value.rb
+++ b/app/models/value.rb
@@ -30,16 +30,27 @@ class Value < ApplicationRecord
       new(row_from_data(data))
     end
 
-    def find_or_create_for_attribute(attribute, name)
+    def find_or_create_for_attribute!(attribute, name)
       friendly_id = generate_friendly_id("#{attribute.friendly_id}__#{name}")
 
-      find_or_create_by(
-        friendly_id: friendly_id,
-      ) do |v|
-        v.name = name
-        v.handle = generate_handle(friendly_id)
-        v.primary_attribute = attribute
-        v.position = attribute.next_position
+      existing_value = find_by(friendly_id: friendly_id)
+
+      return existing_value if existing_value
+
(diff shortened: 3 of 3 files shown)


app/commands/source/add_value_command.rb
```diff
@@ -0,0 +1,99 @@
...
+      required
+    end
+
+    def execute
+      frame("Adding new value") do
+        ActiveRecord::Base.transaction do
+          setup!
+          create_value!
+          create_attributes_value!
+        rescue ActiveRecord::Rollback => e
```
elsom25, line 25: I think its easier to read keeping the exits in method? Saves having to raise simply to catch and exit?

app/models/value_sorter.rb
```diff
@@ -3,8 +3,12 @@
...
   class << self
     def sort(values, locale: "en")
-      if values.first.position.present?
-        values.sort_by(&:position)
+      return values if values.length <= 1
+
+      if values.any? { _1.position.present? }
+        values.sort_by do |v|
+          v.position || values.map(&:position).compact.max + 1
+        end
```
elsom25, line 11: Should this be `values.all?`? That would let us keep the sort simple as well.

When do we have a case of something being manually sorted but missing a position in practice?

app/commands/source/add_value_command.rb
```diff
@@ -0,0 +1,99 @@
...
+
+      raise ActiveRecord::Rollback,
+        "Primary attribute `#{params[:attribute_friendly_id]}` not found" if @primary_attribute.nil?
+
+      @extended_attributes = @primary_attribute.extended_attributes
+      @manually_sorted = @primary_attribute.manually_sorted?
+    end
+
+    def create_value!
+      value_identifier = "#{@primary_attribute.handle}__#{params[:name]}"
```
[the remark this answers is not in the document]
tylerrowsell: +1

app/commands/source/add_value_command.rb
KComrade53: Since there are multiple db inserts to multiple tables, I opted to do it as a transaction and raising a rollback seems to be the common practice for cancelling/failing a transaction. Also it DRYs up the error handling a bit rather than putting `exit(1)` in 4 spots

app/models/value_sorter.rb
```diff
@@ -3,8 +3,12 @@
 module ValueSorter
   class << self
     def sort(values, locale: "en")
-      if values.first.position.present?
-        values.sort_by(&:position)
+      return values if values.length <= 1
```
elsom25, line 6: I wonder if this optimization is worth it? Feels like it adds more complexity for fairly minimal gains? (I expect in the singular case both sorting paths should be fairly minimal overhead)

If we do keep it, one small nit since we generally prefer `size` over `length` in ruby land:
```suggestion
      return values if values.size <= 1
```

lib/string_transformer.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module StringTransformer
```
elsom25, line 3: This name feels _much broader_ than this is in practice? I also think its actually more app-specific than a conventional lib file (lib should be things you can copy-paste into most projects, whereas logic around friendly ids and handles is pretty _domain specific_ for us?)

elsom25: We should also make sure we use this class throughout wherever we're doing stuff in this space of handle/friendly id generation

tylerrowsell (commented):
+1 to what Jesse mentioned. Let's lean into using the models for "create"

app/commands/source/add_value_command.rb
```diff
@@ -0,0 +1,99 @@
...
+      friendly_id = StringTransformer.generate_friendly_id(value_identifier)
+
+      existing_value = Value.find_by(friendly_id: friendly_id)
+
+      raise ActiveRecord::Rollback,
+        "Value `#{friendly_id}` already exists" if existing_value
+
+      position = @primary_attribute.values.map(&:position).compact.max + 1 if @manually_sorted
```
elsom25, line 55: This also feels like it should be handled by the model vs in here?

Also when we write it, I think you can use `max_by`?
```suggestion
      position = @primary_attribute.values.max_by(&:position) + 1 if @manually_sorted
```

app/commands/source/add_value_command.rb
```diff
@@ -0,0 +1,99 @@
...
+    private
+
+    def setup!
+      @primary_attribute = Attribute.find_by(friendly_id: params[:attribute_friendly_id])
+
+      raise ActiveRecord::Rollback,
+        "Primary attribute `#{params[:attribute_friendly_id]}` not found" if @primary_attribute.nil?
+
+      @extended_attributes = @primary_attribute.extended_attributes
+      @manually_sorted = @primary_attribute.manually_sorted?
```
elsom25, line 43: Do we need this ivar or do we just always check against `@primary_attribute.manually_sorted?`?

KComrade53: That's what I had initially but changed it to this so I was just passing around a bool. Happy to change it back since it's not really providing any benefit

app/commands/source/add_value_command.rb
elsom25: But instead you have a raise in 4 spots? Is that actually dryer?

app/commands/source/add_value_command.rb
KComrade53: Working on that as we speak :)

app/models/value_sorter.rb
KComrade53: The changes to this file are no longer needed, so I've reverted them

app/commands/source/add_value_command.rb
```diff
@@ -0,0 +1,80 @@
...
+      desc "Friendly ID of the primary attribute to add the value to"
+      required
+    end
+
+    def execute
+      frame("Adding new value") do
+        ActiveRecord::Base.transaction do
+          setup!
+          create_value!
+          create_attributes_value!
```
tylerrowsell, line 24: Can this be pushed inside of value creation as well? I feel like we did something similar in the parser?

app/commands/source/add_value_command.rb
elsom25: I think the fact this relies on `raise ActiveRecord::Rollback` is what makes this not sit well with me. We're not consolidating, we're adding a layer of redirect + requiring our top-level function to have a wrapped exception handler?

If we decide this pattern is preferred, all commands should be updated to match as well. The goal here is a consistent system. Happy to hash this out IRL to find the right _next best_ pattern

KComrade53: By moving the creation logic out of the script and into the Value model, I was able to remove most of these exception handling, so I switched it back to the log+exit 😄

PR #278, opened 2024-08-19
Add blake3 manifests to generate_dist commands

elsom25:
- In `generate_dist.yml` only consider `data/` files; run the rest based off `@main`
- Downloads (and caches) Blake3 b/c its fast and hot right now
- While we could only compare the keyed manifest, its fast to check both (and the manifest files themselves can be useful for auditing)

### DIFF
--- a/.github/workflows/generate_dist_locales.yml
+++ b/.github/workflows/generate_dist_locales.yml
@@ -16,6 +16,7 @@ permissions:
 
 jobs:
   generate_locales:
+    if: github.repository_owner == 'Shopify'
     runs-on: ubuntu-latest
     outputs:
       LOCALES: ${{ steps.retrieve_locales.outputs.locales }}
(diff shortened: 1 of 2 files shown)


.github/workflows/generate_dist_locales.yml
```diff
@@ -56,7 +81,31 @@ jobs:
...
+        rm -f manifest.b3
+        for manifest in $(ls *_manifest.b3 | sort); do
+          echo "${{ secrets.MANIFEST_KEY }}" | b3sum --keyed - "$manifest" > verify.sig
+          cmp -s "$manifest" verify.sig && cat "$manifest" >> manifest.b3
+        done
+
+        if ! b3sum dist --check manifest.b3; then
+          echo "Error: Integrity check failed. Manifest signature or file integrity mismatch."
+          exit 1
+        fi
```
elsom25, line 108: I think one potential failure here is if there are locales that don't get updated... though conceptually we should be running this against all locales based off the workflow setup

PR #288, opened 2024-08-21
Create 2-localization-fix-requests

ricardotejedorsanz:
Add an issue template to suggest taxonomy localization fixes.

### DIFF
--- a/.github/ISSUE_TEMPLATE/2-localization-fix-requests
+++ b/.github/ISSUE_TEMPLATE/2-localization-fix-requests
@@ -40,7 +40,6 @@ body:
         - Thai
         - Turkish
         - Vietnamese
-        - Other
   - type: dropdown
     id: taxonomy_field
     attributes:


.github/ISSUE_TEMPLATE/2-localization-fix-requests
```diff
@@ -0,0 +1,103 @@
...
+    attributes:
+      label: Are you reporting a Mixed Language Experience (MLX) or a Translation Quality issue?
+      multiple: false
+      options:
+        - Translation Quality issue
+        - MLX (English text in another language)
+  - type: dropdown
+    id: target_language
+    attributes:
+      label: Target Language(s)
```
elsom25, line 19: Do we have stats on language use? Should we come up with some heuristic on what language gets listed vs "othered"?

elsom25 (approved):

.github/ISSUE_TEMPLATE/2-localization-fix-requests
ricardotejedorsanz: @aalexandreia or @MRodriguezRod can confirm, but I believe this matches the languages we support in our customer accounts (excluding beta CAN EN we don't support in the public repo right now):
image

elsom25: @ricardotejedorsanz I guess my mind went to dropping `other` if we have this as exhaustive. But suppose it doesn't hurt to having an escape hatch.

MRodriguezRod: What would be the case scenario where "other" would be selected, given that we have all lang and allow for multiple selections?

ricardotejedorsanz: In that case we can probably drop `other`. This isn't a mandatory field, so users can always leave it blank if a language isn't listed (e.g. suggesting we support another lang). We can always go back to adding `other` later.

PR #297, opened 2024-08-30
update prev shopify taxonomy title

alexmgrant:
Closes: https://github.com/Shopify/temp-project-mover-Archetypically-20250318143024/issues/652

### DIFF
--- a/docs/_layouts/release.html
+++ b/docs/_layouts/release.html
@@ -108,7 +108,7 @@ layout: default
         {% assign rules = mapping.rules %}
         <h3 class="value-title value-title--box mapping-title" tabindex="0">
           {% if output_title == "shopify" %}
-            {{ output_title }} (Previous Versions)
+            {{ output_title }} (Legacy)
           {% else %}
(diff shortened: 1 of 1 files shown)


docs/_layouts/release.html
```diff
@@ -108,7 +108,7 @@ <h2>Channel mappings</h2>
         {% assign rules = mapping.rules %}
         <h3 class="value-title value-title--box mapping-title" tabindex="0">
           {% if output_title == "shopify" %}
-            {{ output_title }} (Previous Versions)
+            {{ output_title }} (Legacy)
```
elsom25, line 111: Fine for now, but what happens when we have a mapping from `2024-07`?

alexmgrant: From what I gathered each _release_ will be mapped to this legacy version (ver 2022-02).

elsom25: Mappings are to the previous version (mapping back to "legacy" would be computed by following mappings release by releaes)

Issue #296, opened 2024-08-28
[Taxonomy Tree]: Stainless steel In jewelery Base Material


jyoti217:
### Taxonomy Field

Values

### Vertical

Apparel & Accessories

### Change Type

New attribute value

### Issue

For Jewelery Base Material Need Stainless steel as an option value in Category metafields.

### Description

For Jewelery Base Material Need Stainless steel as an option value in Category metafields.

PR #325, opened 2024-09-23
Expanding values in [Jewelry material]

ornanatan:
As per https://github.com/Shopify/product-taxonomy/issues/296.


tylerrowsell (approved):
Changes make sense. You're getting a test failure because the localizations haven't been added for other languages yet.

Once we get those (or figure out how we're going to work with argo) we can merge.

PR #341, opened 2024-09-29
[HG] Extend towels category in Home & Garden

ornanatan:
Extend towels category in Home & Garden vertical:
Home & Garden > Linens & Bedding > Towels > Head Towels
Home & Garden > Linens & Bedding > Towels > Hand Towels


ricardotejedorsanz (approved):
All looks good 👍 
We'll need to generate mappings before merging.

PR #355, opened 2024-10-01
[AA][HB] Expand color/pattern attributes in Eyeglasses and similar

ornanatan:
Add color and pattern to eyeglasses


tylerrowsell (approved):
Don't forget to regenerate the dist.

ricardotejedorsanz (changes requested):
I suggest we expand with `Frame color`, `Lens color` and `Temple color` to provide more options, and mirror the same structure for [Sunglasses] (keeping `Color` not to break any flows for existing metadata).

Also, let's add `Lens color` in [Sunglass Lenses] and [Eyeglass Lenses].

image
image
image
image

PR #365, opened 2024-10-04
[BI] Add raw components category in the Construction area

ricardotejedorsanz:
It has been identified that our current taxonomy does not accurately reflect a category for construction component parts that have no use unless assembled with each other to form a working unit. This is particularly crucial for merchants selling these components in Florida, where the surtax differs from that applied to standalone products.

While standalone building material categories exist in the Hardware section (see image below), these could be interpreted as standalone products rather than component parts requiring assembly. Due to the different tax treatment required for these items, we suggest creating a new category with the following path: **Business & Industrial > Construction > Raw Structural Components**. This should more accurately denote construction-related materials such as concrete and wood planks --and could be expanded in the future as needed.

image


ricardotejedorsanz:
cc @drose-shopify, as per our conversation weeks ago — let me know if you have any feedback. 🙏

PR #366, opened 2024-10-04
[AE] Add category for Clothing Sewing Materials

ricardotejedorsanz:
It has been flagged [[issues/3443](https://github.com/Shopify/custom-data-issues/issues/3443)] that our current taxonomy lacks a dedicated category specifically for clothing sewing materials. This gap could potentially affect merchants who need to apply state-specific tax exemptions for items like fabric, thread, buttons, and zippers used in clothing construction, as required by New Jersey and Minnesota laws.

Existing categories under Arts & Entertainment > Hobbies & Creative Arts > Arts & Crafts can be a suitable umbrella for these products, but they do not clearly distinguish between general crafting materials and those intended for clothing (see image below). We propose creating a new category: **Arts & Entertainment > Hobbies & Creative Arts > Arts & Crafts > Clothing Sewing Materials**. This should more accurately denote clothing sewing-related materials and could be expanded in the future as needed or serve as a parent to applicable subcategories such as Craft Fasteners & Closures, Crafting Fibers, or Textiles.

image
image

### DIFF
--- a/data/categories/ae_arts_entertainment.yml
+++ b/data/categories/ae_arts_entertainment.yml
@@ -35,6 +35,7 @@
   - ae-2-1-4
   - ae-2-1-5
   - ae-2-1-6
+  - ae-2-1-7
   attributes: []
 - id: ae-2-1-1
   name: Art & Craft Kits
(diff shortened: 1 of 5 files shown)


ricardotejedorsanz:
fyi @FredyPla, as per this issue: https://github.com/Shopify/custom-data-issues/issues/3443

tylerrowsell (approved):
Looks good from a technical side of things.

I trust your judgement here, but generally we should be careful adding things simply because it doesn't fit into a tax category. We don't want our taxonomy system to be become overcomplicated with exceptions.

In this case it seems like it can fit though.

PR #367, opened 2024-10-04
[EL] Add design attribute to Microphones

ricardotejedorsanz:
As per [https://github.com/Shopify/custom-data-issues/issues/3230](https://github.com/Shopify/custom-data-issues/issues/3230).

Adding a new attribute in **Electronics > Audio > Audio Components > Microphones** to account for different designs:

image
image
image


data/attributes.yml
```diff
@@ -19021,13 +19021,13 @@ base_attributes:
   - language_version__cro_croatian
   - language_version__cym_welsh
   - language_version__dan_danish
+  - language_version__deu_german
   - language_version__deu_be_german_belgium
   - language_version__deu_ch_german_switzerland
-  - language_version__deu_german
```
tylerrowsell, line 19026: Just seems like some lines moving around here? Not a problem but just want to note it in case something funny is going on and you're getting things from another PR?

tylerrowsell (approved):

Issue #262, opened 2024-08-05
[Taxonomy Tree]: Micrometers


victorlaforga:
### Change Type

None

### Issue

Add Micrometers and specific machining parts and elements like clamping heads

### Description

See

PR #377, opened 2024-10-06
[HA] Add Micrometers and Micrometer Accessories

ricardotejedorsanz:
In response to https://github.com/Shopify/product-taxonomy/issues/262

Here are some examples of these products sold by our merchants:

image
image
image

### DIFF
--- a/data/categories/ha_hardware.yml
+++ b/data/categories/ha_hardware.yml
@@ -8423,6 +8423,7 @@
   - color
   - device_technology
   - display_technology
+  - micrometer_type
   - pattern
   - system_of_measurement
 - id: ha-15-37


data/categories/ha_hardware.yml
```diff
@@ -8408,6 +8416,15 @@
   - display_technology
   - measuring_tool_features
   - pattern
+- id: ha-15-36-39
+  name: Micrometers
```
ornanatan, line 8420: I think we need to add a type attribute as well for Outside Micrometer, Digital Micrometer, Inside Micrometer, Depth Micrometer, Indicating Micrometer, Bore Micrometer, Blade Micrometer, Screw Thread Micrometer, Disc Micrometer, Groove Micrometer

ricardotejedorsanz: Added! ✅

tylerrowsell (approved):

PR #378, opened 2024-10-06
[FB] Rename Dry Beans as Dry Legumes

ricardotejedorsanz:
It has been flagged that we're missing an appropriate category for Lentils, but our tree does provide a suitable category under Food, Beverages & Tobacco > Food Items > Fruits & Vegetables > Dry Beans. This includes the attribute for dry bean varieties like lentils, chickpeas, and soybeans.

However, the current names can be confusing. Renaming it to a broader term like "dry legumes" would clarify that other products considered pulses also have a suitable match here.


ricardotejedorsanz:
@tylerrowsell Just bringing to your attention that this includes a renamed attribute and category (with no substantial change), which we should check when importing to Core.

MartinEconea:
Oooo! Thank you!

data/attributes.yml
```diff
@@ -5042,24 +5042,24 @@ base_attributes:
   - nut_butter_variety__pistachio
   - nut_butter_variety__walnut
 - id: 1492
-  name: Dry bean variety
-  description: Identifies the variety of dry beans used, e.g. chickpeas, black beans
-  friendly_id: dry_bean_variety
```
tylerrowsell, line 5047: We definitely can't change this handle without a mapping. As merchants will have metafields already based on this.

ricardotejedorsanz:
Addressed in https://github.com/Shopify/product-taxonomy/pull/745 to avoid version bumps. Category rename only, attribute handle stays as is.

PR #380, opened 2024-10-06
[AE] Add Mesh type attribute to Fabrics area

ricardotejedorsanz:
It has been noted (https://github.com/Shopify/product-taxonomy/issues/223) that we lack specificity in mesh fabrics, such as airtex or athletic fabrics. While there's an argument for creating a separate subcategory, our current structure distinguishes fabrics by material rather than structure.

For example, certain mesh types apply to cotton, which would create unnecessary duplications in the tree. We suggest introducing a new attribute, "mesh type," applicable to the parent category and cotton.

image
image
image
image
image


ornanatan (approved):
Looks great

PR #406, opened 2024-10-15
[BT] Reparent categories under Feeding Essentials

ricardotejedorsanz:
For consistency, reparenting under `Feeding Essentials` the following categories and subcategories: `Baby Bottle Nipples & Liners`, `Baby Bottles`, `Bibs` and `Sippy Cups`.

@chesterbot01 @jeanrick This is the first example of mappings across more than one Shopify legacy version. I would appreciate your reviews to ensure everything is as expected.

image
image

### DIFF
--- a/data/categories/bt_baby_toddler.yml
+++ b/data/categories/bt_baby_toddler.yml
@@ -1818,10 +1818,7 @@
   name: Nursing & Feeding
   children:
   - bt-10-1
-  - bt-10-2
-  - bt-10-3
   - bt-10-4
-  - bt-10-5
   - bt-10-6
   - bt-10-7
   - bt-10-8
@@ -1831,7 +1828,6 @@
   - bt-10-12
   - bt-10-13
   - bt-10-14
-  - bt-10-15
   - bt-10-16
   attributes: []
 - id: bt-10-1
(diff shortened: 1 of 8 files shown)


dist/en/integrations/shopify/shopify_2024-07_to_shopify_2024-10.txt
```diff
@@ -0,0 +1,4 @@
+# Shopify Product Taxonomy - Mapping shopify/2024-07 to shopify/2024-10-unstable
+# Format:
+# → {base taxonomy category name}
+# ⇒ {mapped taxonomy category name}
```
tylerrowsell, line 4: Something seems off here... there shouldn't have been mappings here?

dist/en/integrations/shopify/shopify_2024-07_to_shopify_2024-10.json
```diff
@@ -0,0 +1,217 @@
...
+
+      ]
+    },
+    {
+      "input_taxonomy": "shopify/2022-02",
+      "output_taxonomy": "shopify/2024-10-unstable",
+      "rules": [
+
+      ]
+    },
```
tylerrowsell, line 17: Something things wrong with the generate mappings command. I wouldn't expect other mappings to show in the file specific to 2024-07 -> 2024-10

@chesterbot01

dist/en/integrations/shopify/shopify_2024-07_to_shopify_2024-10.txt
ricardotejedorsanz: @chesterbot01

tylerrowsell:
>  is there a quick way to bypass the async localization step by reusing existing translations and fit this change into the 2024-10 release before we freeze an RC?

You could update the values in each locale file in `/data`

Likely something chat gpt can help you write a script for.

```
For each file in the `localizations/categories/*.yml directory update the keys that start with bt-10-2 to bt-10-16-5 
```

ricardotejedorsanz:
> You could update the values in each locale file in `/data`
> Likely something chat gpt can help you write a script for.

@tylerrowsell The script works well to replace the old ids with the new ones in `/data`, but the parent-id `bt-10-16` is a new category that only exists in the `en` locale... which blocks the generation of the `.dist` files. We might need to postpone to 2025-01, unless I'm missing something?

ornanatan (approved):

PR #426, opened 2024-10-29
[SE] Expanding categories in services vertical

ornanatan:
The merchant sells Language Courses and there is not education category in the tree that can be fit for his need
https://merchant-frustrations.shopifycloud.com/features/12672?page=5#ticket-446054
Tax services also a service that was missing from the tree.
The annotators found that Waste Management Services were missing from the tree, i created a category named Environmental Services and located it as sub category.
https://shopify.slack.com/archives/C07AH90BQ7Q/p1728650752963499

### DIFF
--- a/data/localizations/categories/en.yml
+++ b/data/localizations/categories/en.yml
@@ -24249,11 +24249,8 @@ en:
       name: Property Management
       context: Services > Professional Services > Property Management
     se-7-9:
-      name: Educational Services
-      context: Services > Professional Services > Educational Services
-    se-7-10:
-      name: Tax Exemption Services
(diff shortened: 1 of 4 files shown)


data/localizations/categories/en.yml
```diff
@@ -24248,6 +24248,18 @@ en:
     se-7-8:
       name: Property Management
       context: Services > Professional Services > Property Management
+    se-7-9:
+      name: Educational Services
+      context: Services > Professional Services > Educational Services
```
ricardotejedorsanz, line 24253: We have this under `se-6-2`, no need to duplicate.

PR #437, opened 2024-11-08
Add Sorbet

danielpgross:
Sets up the project to use Sorbet.

I hesitated at first to bring in Sorbet out of concern that it might make the repo less accessible to external developers who aren't familiar. But then I found myself writing YARD docs on method parameter types and thinking about how to handle runtime validation of arguments.

Sorbet does both well (and more), so let's run with it.

PR #438, opened 2024-11-11
Revert "Add Sorbet"

danielpgross:
Reverts Shopify/product-taxonomy#437

Considering this more, Sorbet seems too heavy-handed for a small app like this one. We will handle documenting method param types using YARD (as needed) and runtime validation of method args manually in the places where it really matters (i.e. consuming input data).


elsom25 (approved):

elsom25:
It was worth trying but agreed feels to heavy. At least until we get much stronger evidence it simplifies a lot for us

PR #441, opened 2024-11-13
[VP] Expanding categories in the Aircraft area

ornanatan:
As part of expanding V&P vertical, some of the categories in air craft need expansion.

image
image


data/categories/vp_vehicles_parts.yml
```diff
@@ -4453,6 +4453,160 @@
...
+  - vp-2-1-2-3
+  attributes:
+  - body_color
+  - body_pattern
+  - fuel_supply
+  - item_condition
+  - upholstery_color
+  - upholstery_pattern
+- id: vp-2-1-2-1
+  name: Light Jets
```
ricardotejedorsanz, line 4514: It seems the terms "private jet" and "business jet" can be used interchangeably (https://en.wikipedia.org/wiki/Business_jet). I suggest we call the parent `Business Jets` instead of "private jets" to avoid focusing on the actual ownership, which can vary without defining the product, and keep two child nodes instead of three: `Light Jets` and `Turboprop Aircraft`.

data/categories/vp_vehicles_parts.yml
```diff
@@ -4453,6 +4453,160 @@
...
+  - upholstery_pattern
+- id: vp-2-1-2-2
+  name: Turboprop Aircraft
+  children: []
+  attributes:
+  - body_color
+  - body_pattern
+  - fuel_supply
+  - item_condition
+  - upholstery_color
```
ricardotejedorsanz, line 4531: Actually, should we also add `Very Light Jets` under `vp-2-1-2`?
https://en.wikipedia.org/wiki/Very_light_jet

ricardotejedorsanz (approved):
I agree these changes make sense if we are aiming for product exhaustivity, with a couple of minor suggestions I would address before merging. Drones in particular makes me think it might be worth exploring a robotics section as well (as we currently only have robotic toys in the Toys & Games vertical).

ornanatan:
@alokaraju 
 Thank you for the suggestions. Regarding the condition, we already have 'Item Condition' attributes that cover new, used, and refurbished statuses. The size/capacity will be included in the unit of measurement attributes that @ricardotejedorsanz is currently working on. As for the usage, it can indeed be inferred from the category names as they are clearly defined

data/categories/vp_vehicles_parts.yml
```diff
@@ -4453,6 +4453,160 @@
...
+  - upholstery_pattern
+- id: vp-2-1-2
+  name: Private Jets
+  children:
+  - vp-2-1-2-1
+  - vp-2-1-2-2
```
danielpgross, line 4504: I don't think it makes sense for `Turboprop Aircraft` to be a child of `Private Jets`:
* Turboprops aren't jets by definition (different engine type)
* A turboprop can also fall under `Commercial Aircraft` (e.g. a Q400 passenger plane)

danielpgross (commented):
We should definitely add a category here for light general aviation (e.g. Cessna 172), as that will probably be one of the more commonly used ones. We could also add ultralights, hot air balloons, and hang gliders/paragliders.

We should also move engine type into an attribute (jet, turboprop, piston prop) and update the categories accordingly (e.g. rename `Private Jets`)

tylerrowsell (approved):

danielpgross (changes requested):
Left some comments, including a typo.

Not blocking, but we could also consider adding more attributes, for example:

**Cockpit type**

- Analog
- Glass
- Hybrid
- Integrated

**Safety features**

- Terrain Awareness and Warning System (TAWS)
- Traffic Collision Avoidance System (TCAS)
- Emergency Locator Transmitter (ELT)
- Stall Warning System
- Parachute System

ricardotejedorsanz:
@alokaraju @danielpgross The PR is ready to be merged. Unless you have any blocking unaddressed concerns, can you please approve your change requests? cc @ornanatan

data/localizations/categories/en.yml
```diff
@@ -31040,6 +31040,54 @@ en:
     vp-2-1:
       name: Aircraft
       context: Vehicles & Parts > Vehicles > Aircraft
+    vp-2-1-1:
+      name: Heavier-Than-Air Aircraft
+      context: Vehicles & Parts > Vehicles > Aircraft > Heavier-Than-Air Aircraft
```
danielpgross, line 31045: I'm not sure if I agree with having `Heavier-Than-Air Aircraft` as its own category 🤔 It seems unlikely that someone would ever pick it instead of one of the subcategories.

I think I would remove `Heavier-Than-Air Aircraft`and `Lighter-Than-Air Aircraft` and just move all the subcategories to be directly under `Aircraft`.

danielpgross (commented):
Thanks for addressing all the previous feedback - looks good. I just flagged one remaining concern.

PR #443, opened 2024-11-14
Add `Attribute` loading from YAML source with validation

danielpgross:
Adds `Attribute#load_from_source` for loading from source YAML, reusing the `ModelIndex` and validation work introduced in the previous PR (#439).

The new challenges here were:

- Depending on previously loaded (`Value`) data in the method. This is done by passing in a hash from the `ModelIndex` instance for fast lookup.
- Resolving references to `Value` instances in the `Attribute` model and `Attribute` instances in the `ExtendedAttribute` model. I wanted this to play nicely with the existing ActiveModel validation pattern, so I opted for attempted resolution of the model object and falling back to just passing the `friendly_id` if the object couldn't be found. This makes it simple to implement validation with ActiveModel.
- Handling a more complex YAML input format, with 2 required keys, each containing an array of values with a different format (base vs. extended).

### DIFF
--- a/dev/lib/product_taxonomy/models/attribute.rb
+++ b/dev/lib/product_taxonomy/models/attribute.rb
@@ -91,7 +91,11 @@ module ProductTaxonomy
 
     def values_valid?
       values&.each do |value|
-        errors.add(:values, "Value with friendly ID \"#{value}\" was not found") unless value.is_a?(Value)
+        errors.add(
+          :values,
+          :not_found,
+          message: "could not be resolved for friendly ID \"#{value}\"",
+        ) unless value.is_a?(Value)
       end
     end
   end
--- a/dev/test/models/attribute_test.rb
+++ b/dev/test/models/attribute_test.rb
@@ -88,9 +88,16 @@ module ProductTaxonomy
         extended_attributes: []
       YAML
 
-      assert_raises(ActiveModel::ValidationError) do
+      error = assert_raises(ActiveModel::ValidationError) do
         Attribute.load_from_source(source_data: YAML.safe_load(yaml_content), values: {})
       end
+      expected_errors = {
+        friendly_id: [{ error: :blank }],
+        handle: [{ error: :blank }],
+        description: [{ error: :blank }],
+        values: [{ error: :blank }],
+      }
+      assert_equal expected_errors, error.model.errors.details
     end
 
     test "load_from_source raises an error if the source data contains attributes with empty values" do
@@ -106,9 +113,13 @@ module ProductTaxonomy
         extended_attributes: []
       YAML
 
-      assert_raises(ActiveModel::ValidationError) do
+      error = assert_raises(ActiveModel::ValidationError) do
         Attribute.load_from_source(source_data: YAML.safe_load(yaml_content), values: {})
       end
+      expected_errors = {
+        values: [{ error: :blank }],
+      }
+      assert_equal expected_errors, error.model.errors.details
     end
 
     test "load_from_source raises an error if the source data contains attributes with values that are not found" do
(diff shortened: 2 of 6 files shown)


dev/test/models/attribute_test.rb
```diff
@@ -0,0 +1,235 @@
...
+        ---
+        base_attributes:
+        - id: 1
+          name: Color
+        extended_attributes: []
+      YAML
+
+      assert_raises(ActiveModel::ValidationError) do
+        Attribute.load_from_source(source_data: YAML.safe_load(yaml_content), values: {})
+      end
```
tylerrowsell, line 93: Should these check the error message to ensure it is not a false positive?

tylerrowsell (approved):

dev/test/models/attribute_test.rb
danielpgross: Yeah good call, I was avoiding this because I thought I'd need to assert against the exact error message, which would make the tests brittle. But it looks like I can set an error key and assert against that instead, which means the exact error message wording can still change. I'm going to try that.

dev/lib/product_taxonomy/cli.rb
```diff
@@ -5,7 +5,13 @@ class Cli < Thor
     desc "dist", "Generate the taxonomy distribution"
     def dist
       seconds = Benchmark.realtime do
-        Value.load_from_source(source_data: YAML.safe_load_file("../data/values.yml"))
+        source_data = YAML.safe_load_file("../data/attributes.yml")
```
elsom25, line 8: Minor nitpick:

Why do we load this before the line below? It made me think we need this before we load values.

Also should we give this a different name since we're now loading multiple "models"?

dev/lib/product_taxonomy/models/attribute.rb
```diff
@@ -2,15 +2,101 @@
...
+      # @return [ModelIndex<Attribute>] A model index of {Attribute} objects.
+      def load_from_source(source_data:, values:)
+        model_index = ModelIndex.new(self, hashed_by: :friendly_id)
+        raise ArgumentError, "source_data must be a hash" unless source_data.is_a?(Hash)
+        raise ArgumentError, "source_data must contain keys \"base_attributes\" and \"extended_attributes\"" unless
+          source_data.keys.sort == ["base_attributes", "extended_attributes"]
+
+        ["base_attributes", "extended_attributes"].each do |type|
+          load_attributes(type, source_data, model_index, values)
+        end
```
elsom25, line 21: Why this vs a more plain:
```rb
load_attributes("base_attributes", source_data, model_index, values)
load_attributes("extended_attributes", source_data, model_index, values)
```

dev/lib/product_taxonomy/models/attribute.rb
```diff
@@ -2,15 +2,101 @@
...
+        model_index
+      end
+
+      private
+
+      def load_attributes(type, source_data, model_index, values)
+        source_data[type].each do |attribute_data|
+          raise ArgumentError, "source_data must contain hashes" unless attribute_data.is_a?(Hash)
+
+          attribute = if type == "base_attributes"
```
elsom25, line 32: Given this conditional, was it worth splitting out method?

dev/lib/product_taxonomy/models/attribute.rb
```diff
@@ -2,15 +2,101 @@
...
+
+    private
+
+    def values_valid?
+      values&.each do |value|
+        errors.add(
+          :values,
+          :not_found,
+          message: "could not be resolved for friendly ID \"#{value}\"",
+        ) unless value.is_a?(Value)
```
elsom25, line 98: nit: trailing conditionals on a block are easy to miss:

```rb
next if value.is_a?(Value)

errors.add(...)
```

dev/lib/product_taxonomy/models/attribute.rb
danielpgross: Agreed, on second thought the block is unnecessary here

dev/lib/product_taxonomy/cli.rb
danielpgross: You're right. I'm using this command as an easy entrypoint into the program for now, mainly so I can keep an eye on performance/runtime. Until the `dist` command is actually implemented, whatever is here can be considered scratch work/temporary.

But maybe it would be less confusing to move it somewhere else, like a test case without any assertions that just measures runtime.

PR #444, opened 2024-11-14
Refactor and move benchmarking code

danielpgross:
Refactor to address feedback in #443. The most notable change is the introduction of `benchmarks.rb` for dedicated benchmarking (which was previously being done in a scrappy way in `cli.rb`). We can use this to keep an eye on runtime performance as we continue to build.

### DIFF
--- a/dev/lib/product_taxonomy/cli.rb
+++ b/dev/lib/product_taxonomy/cli.rb
@@ -4,16 +4,7 @@ module ProductTaxonomy
   class Cli < Thor
     desc "dist", "Generate the taxonomy distribution"
     def dist
-      seconds = Benchmark.realtime do
-        source_data = YAML.safe_load_file("../data/attributes.yml")
-        values_model_index = Value.load_from_source(source_data: YAML.safe_load_file("../data/values.yml"))
-
-        Attribute.load_from_source(
-          source_data:,
-          values: values_model_index.hashed_by(:friendly_id),
-        )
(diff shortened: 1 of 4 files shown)


dev/lib/product_taxonomy/cli.rb
```diff
@@ -4,16 +4,7 @@ module ProductTaxonomy
...
     desc "dist", "Generate the taxonomy distribution"
     def dist
-      seconds = Benchmark.realtime do
-        source_data = YAML.safe_load_file("../data/attributes.yml")
-        values_model_index = Value.load_from_source(source_data: YAML.safe_load_file("../data/values.yml"))
-
-        Attribute.load_from_source(
-          source_data:,
-          values: values_model_index.hashed_by(:friendly_id),
-        )
```
elsom25, line 14: Did we want to keep this part? Since it's effective the dist generation bits right? Or I guess not quite yet

elsom25 (approved):

dev/lib/product_taxonomy/cli.rb
danielpgross: Not quite yet, and the final version will likely be a single call into `Taxonomy.load_from_source` ([like in the prototype](https://github.com/Shopify/product-taxonomy/blob/dg/simplify/lib/product_taxonomy/cli.rb#L8)) rather than loading the different models here.

PR #445, opened 2024-11-17
[VP] Add categories for Motor Sports Vehicles

ornanatan:
Adding new categories for motor sports vehicles.
Create a new attribute for Tire application type.
Please note that attributes like Engine Power (HP),Fuel Capacity, Top Speed, Horsepower etc will be add as part as the unit of measurement task


data/categories/vp_vehicles_parts.yml
```diff
@@ -4765,6 +4766,150 @@
   - upholstery_color
   - upholstery_pattern
   - vehicle_features
+- id: vp-2-2-7
+  name: Motor Sports Vehicles
```
alokaraju, line 4770: @ornanatan how about monster trucks?

data/localizations/categories/en.yml
```diff
@@ -31082,6 +31082,39 @@ en:
     vp-2-2-6:
       name: Snowmobiles
       context: Vehicles & Parts > Vehicles > Motor Vehicles > Snowmobiles
+    vp-2-2-7:
+      name: Motor Sports Vehicles
+      context: Vehicles & Parts > Vehicles > Motor Vehicles > Motor Sports Vehicles
+    vp-2-2-7-1:
+      name: Race Cars
+      context: Vehicles & Parts > Vehicles > Motor Vehicles > Motor Sports Vehicles > Race Cars
```
ricardotejedorsanz, line 31090: Categories make sense to me. I suggest we add a few additional leaves like GT Cars, Stock Cars or Racing Trucks.

PR #450, opened 2024-11-18
Refactor ModelIndex to use hashes instead of sets

danielpgross:
Previously, ModelIndex used `Set`s to track values already seen, and **could only check uniqueness on a model _before_ it was added to the index**. With this PR, uniqueness can be checked **either before or after the model is added to the index**. This was necessary for category loading, since we need to add the model to the index as part of building the tree relationships, but can't do validation until those tree relationships are complete. This change actually simplifies ModelIndex; it now only uses hashes (not sets) and reuses the same hashes for both uniqueness checks and fast lookups.


elsom25 (approved):
I like the ergonomics change that comes with this 💅

PR #452, opened 2024-11-19
Add localization for model attributes

danielpgross:
Adds localization support for attributes, categories, and values through a new `Localized` mixin. We'll need this for generating the full, localized output for distribution.

- Created a new `Localized` mixin that handles loading translations from YAML files
- Integrated localization support into `Attribute`, `Category`, and `Value` models
- Added `mocha` gem for testing mocks and stubs (was needed for tests covering loading localization YAML)

The approach builds on the current/previous one (ex. [here](https://github.com/Shopify/product-taxonomy/blob/main/app/models/attribute.rb#L85)). I tried to improve on it a bit by pulling the shared code into a mixin (`Localized`) with some metaprogramming. This kind of metaprogramming is sometimes considered a code smell because it can be harder to follow. But I think it makes sense to use here, since we don't use Sorbet and the codebase is small.

### DIFF
--- a/dev/lib/product_taxonomy/models/attribute.rb
+++ b/dev/lib/product_taxonomy/models/attribute.rb
@@ -3,6 +3,7 @@
 module ProductTaxonomy
   class Attribute
     include ActiveModel::Validations
+    extend Localized
 
     class << self
       # Load attributes from source data. By default, this data is deserialized from a YAML file in the `data` directory.
@@ -69,7 +70,9 @@ module ProductTaxonomy
     validates_with ProductTaxonomy::ModelIndex::UniquenessValidator, attributes: [:friendly_id]
     validate :values_valid?
 
-    attr_reader :id, :name, :description, :friendly_id, :handle, :values, :uniqueness_context
+    localized_attr_reader :name, :description
+
+    attr_reader :id, :friendly_id, :handle, :values, :uniqueness_context
 
     # @param id [Integer] The ID of the attribute.
     # @param name [String] The name of the attribute.
--- a/dev/lib/product_taxonomy/models/mixins/localized.rb
+++ b/dev/lib/product_taxonomy/models/mixins/localized.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+module ProductTaxonomy
+  module Localized
+    def localizations
+      return @localizations if defined?(@localizations)
+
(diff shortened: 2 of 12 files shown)


danielpgross:
* **#452** <a href="https://app.graphite.dev/github/pr/Shopify/product-taxonomy/452?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/product-taxonomy/452?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

tylerrowsell (approved):

dev/lib/product_taxonomy/models/mixins/localized.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module ProductTaxonomy
+  module Localized
+    def localizations
```
elsom25, line 5: Should this be private? I suppose not strictly required, but I usually try to minimize method polution when metaprogramming

dev/lib/product_taxonomy/models/attribute.rb
```diff
@@ -69,7 +70,10 @@ def extended_attribute_from(attribute_data:, model_index:)
     validates_with ProductTaxonomy::ModelIndex::UniquenessValidator, attributes: [:friendly_id]
     validate :values_valid?
 
-    attr_reader :id, :name, :description, :friendly_id, :handle, :values, :uniqueness_context
+    localizations_keyed_by :friendly_id
+    localized_attr_reader :name, :description
```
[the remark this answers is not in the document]
elsom25: This also means we don't have to persist this data (and conceptually, you coul dhave different attributes keyed by differnt things so more flexible 😉 )

elsom25 (approved):
A few stylistic nits, but I like it 🚀

dev/lib/product_taxonomy/models/mixins/localized.rb
danielpgross: True, will make them private

danielpgross:
### Merge activity

* **Nov 20, 12:26 PM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Shopify/product-taxonomy/452).

PR #453, opened 2024-11-20
[VP] Expand protective gear area with Motor Sports

ornanatan:
Rename Vehicles & Parts > Vehicle Parts & Accessories > Vehicle Safety & Security > Off-Road & All-Terrain Vehicle Protective Gear to Vehicles & Parts > Vehicle Parts & Accessories > Vehicle Safety & Security > Off-Road & **Motor Sports** Vehicle Protective Gear
I didn't archive it but create since the rename doesn't effect the meaning of the category.

### DIFF
--- a/data/categories/vp_vehicles_parts.yml
+++ b/data/categories/vp_vehicles_parts.yml
@@ -3817,11 +3817,11 @@
   name: Vehicle Safety & Security
   children:
   - vp-1-6-1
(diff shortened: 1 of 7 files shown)


tylerrowsell:
IMO we need to archive and create a new one here. Changing the name will break CSV imports.

ricardotejedorsanz (approved):

PR #465, opened 2024-11-26
Fix memoization logic in `Localized` mixin

danielpgross:
Right now, the memoization logic will return `nil` if `@localizations` was set to `nil`. This means we don't have a straightforward way to reset this class-level variable. It will be important in subsequent PRs to be able to do so when running tests.

If the localizations are empty, `@localizations` will anyway be set to `{}`, not `nil`. This PR ensures that the localizations are loaded instead of early returning when `@localizations == nil`.

### DIFF
--- a/dev/lib/product_taxonomy/models/mixins/localized.rb
+++ b/dev/lib/product_taxonomy/models/mixins/localized.rb
@@ -3,7 +3,7 @@
 module ProductTaxonomy
   module Localized
     def localizations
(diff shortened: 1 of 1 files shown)


danielpgross:
### Merge activity

* **Nov 27, 1:41 PM EST**: [Graphite](https://app.graphite.dev/github/pr/Shopify/product-taxonomy/465) rebased this pull request as part of a merge.

PR #466, opened 2024-11-26
Add `dist` serialization to `Value` model

danielpgross:
Adds JSON and TXT serialization to the `Value` model for `dist` generation, including methods for both individual values and collections.

- The TXT format uses padding based on the longest GID for consistent alignment.
  - The longest GID is found by searching the index hash, rather than iterating over all `Value` objects.
  - This approach proved to take only a few milliseconds vs. ~20 milliseconds to iterate over all `Value` objects on my machine.
  - Since we do this for each of 21 locales, that difference actually becomes significant.
- Adds back the `hashed_by` method to the `Indexed` mixin to retrieve the full hash of models indexed by a given field.

### DIFF
--- a/dev/lib/product_taxonomy/models/mixins/indexed.rb
+++ b/dev/lib/product_taxonomy/models/mixins/indexed.rb
@@ -47,6 +47,16 @@ module ProductTaxonomy
       hashed_models[field][value]
     end
 
+    # Get the hash of models indexed by a given field. Only works for fields marked unique.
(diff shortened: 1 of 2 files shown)


danielpgross:
### Merge activity

* **Nov 27, 2:06 PM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Shopify/product-taxonomy/466).

PR #467, opened 2024-11-26
Add `dist` serialization to `Attribute` model

danielpgross:
Adds JSON and TXT serialization to the `Attribute` model for `dist` generation, including methods for both individual values and collections.

- Adds a reference from a (base) Attribute to all ExtendedAttributes that refer to it, used for JSON serialization
- TXT padding is calculated based on longest GID, which is implemented simply by iterating through the collection. This only took a few milliseconds on my machine, so is fine for now.
- Also fixes broken localization behaviour for `ExtendedAttribute` that went unnoticed before. I've added tests for `ExtendedAttribute` to ensure that doesn't happen again.

### DIFF
--- a/dev/lib/product_taxonomy/models/extended_attribute.rb
+++ b/dev/lib/product_taxonomy/models/extended_attribute.rb
@@ -2,6 +2,12 @@
 
 module ProductTaxonomy
   class ExtendedAttribute < Attribute
+    class << self
+      def localizations
+        superclass.localizations # Extended attribute localizations are defined in the same place as attributes
+      end
+    end
+
     validate :values_from_valid?
 
     attr_reader :values_from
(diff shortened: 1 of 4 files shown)


dev/lib/product_taxonomy/models/extended_attribute.rb
```diff
@@ -14,6 +20,7 @@ class ExtendedAttribute < Attribute
     #   instead.
     def initialize(name:, handle:, description:, friendly_id:, values_from:)
       @values_from = values_from
+      values_from.add_extended_attribute(self) if values_from.is_a?(Attribute)
```
elsom25, line 23: I would have expected this to be `if values_from.is_a?(ExtendedAttribute)` based on other code

elsom25 (approved):

PR #478, opened 2024-12-03
[AE] Expand string instruments categories

ornanatan:
Expand the string instruments category: Banjos, Mandolins, Ukuleles


data/categories/ae_arts_entertainment.yml
```diff
@@ -7740,6 +7743,201 @@
...
+- id: ae-2-8-7-7-4
+  name: Tenor Banjos
+  children: []
+  attributes:
+  - instrument_color
+  - instrument_material
+  - instrument_pattern
+  - instrument_string_color
+  - instrument_string_material
+  - instrument_string_pattern
```
ricardotejedorsanz, line 7799: Tenor banjos typically have four strings, so I would skip `Tenor Banjos` and keep `4-String Banjos`, `5-String Banjos` and `6-String Banjos` to avoid overlaps.
image

data/categories/ae_arts_entertainment.yml
```diff
@@ -7740,6 +7743,201 @@
...
+- id: ae-2-8-7-8-3
+  name: Bowlback Mandolins
+  children: []
+  attributes:
+  - instrument_color
+  - instrument_material
+  - instrument_pattern
+  - instrument_string_color
+  - instrument_string_material
+  - instrument_string_pattern
```
ricardotejedorsanz, line 7843: Since bowl-back refers to the back shape and both A-style and F-style mandolins could have that back shape, I would skip `Bowlback Mandolins` as a category and instead add an attribute for `Mandolin back shape`, including values like Bowl-back, Flat-back and Archtop -- to avoid overlaps.
image

ricardotejedorsanz (approved):
Code looks great. I would two categories to avoid overlaps as per the comments below.

PR #469, opened 2024-11-26
Add `dist` command to generate distribution files

danielpgross:
Adds CLI command infrastructure and implements the `dist` command for generating taxonomy distribution files.

- Creates a base `Command` class with logging and benchmarking capabilities
  - Adds quiet and verbose logging options for all commands
- Adds `GenerateDistCommand` to handle taxonomy distribution generation
  - CLI options for version and locale selection
  - support for generating taxonomy files in txt/json for all locales
- `GenerateDistCommand` isn't well suited to unit tests, it should be covered with integration tests. I've left that for a follow-up PR since there are some questions about the best way to do it.

Example runs:
```
❯ bin/product_taxonomy dist                  
Version: 2025-01-unstable
Locales: en
Generating files for en
Completed in 1.18 seconds
```

```
❯ bin/product_taxonomy dist --locales=all
Version: 2025-01-unstable
Locales: cs, da, de, en, es, fi, fr, it, ja, ko, nb, nl, pl, pt-BR, pt-PT, sv, th, tr, vi, zh-CN, zh-TW
Generating files for cs
Generating files for da
Generating files for de
Generating files for en
Generating files for es
Generating files for fi
Generating files for fr
Generating files for it
Generating files for ja
Generating files for ko
Generating files for nb
Generating files for nl
Generating files for pl
Generating files for pt-BR
Generating files for pt-PT
Generating files for sv
Generating files for th
Generating files for tr
Generating files for vi
Generating files for zh-CN
Generating files for zh-TW
Completed in 17.08 seconds
```

PR #480, opened 2024-12-04
Add `dist` generation of integrations `all_mappings.json`

danielpgross:
Adds support for generating a consolidated `all_mappings.json` file containing all taxonomy mappings across integrations. This file is present in the old/current `dist` generation, but wasn't included in https://github.com/Shopify/product-taxonomy/pull/469.

To maintain performance, the `IntegrationVersion#to_json` return value is memoized the first time, so it doesn't need to be recomputed to generate `all_mappings.json`

A few additional improvements:
* Current shopify version is passed in from CLI arguments, instead of being read from the hardcoded value in the `VERSION` file
* Output path is defined in one place instead of being repeated

### DIFF
--- a/dev/lib/product_taxonomy/models/integration_version.rb
+++ b/dev/lib/product_taxonomy/models/integration_version.rb
@@ -11,19 +11,28 @@ module ProductTaxonomy
       # Generate all distribution files for all integration versions.
       #
       # @param output_path [String] The path to the output directory.
-      def generate_all_distributions(output_path:, logger:, base_path: nil)
-        load_all_from_source(base_path:).each { _1.generate_distributions(output_path:, logger:) }
+      # @param logger [Logger] The logger to use for logging messages.
+      # @param current_shopify_version [String] The current version of the Shopify taxonomy.
+      # @param base_path [String] The path to the base directory containing integration versions.
+      def generate_all_distributions(output_path:, logger:, current_shopify_version:, base_path: nil)
+        integration_versions = load_all_from_source(current_shopify_version:, base_path:)
+        all_mappings = integration_versions.each_with_object([]) do |integration_version, all_mappings|
+          logger.info("Generating integration mappings for #{integration_version.name}/#{integration_version.version}")
+          integration_version.generate_distributions(output_path:)
+          all_mappings.concat(integration_version.to_json(direction: :both))
+        end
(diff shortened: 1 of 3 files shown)


dev/lib/product_taxonomy/models/integration_version.rb
```diff
@@ -54,25 +61,48 @@ def load_from_source(integration_path:)
...
+        File.write(
+          File.expand_path("en/integrations/all_mappings.json", output_path),
+          JSON.pretty_generate(to_json(mappings:, version:)),
         )
       end
+
+      def to_json(version:, mappings:)
+        {
+          version:,
+          mappings: mappings.is_a?(Array) ? mappings : [mappings],
```
elsom25, line 78: Nit: The array constructor does the same thing:
```suggestion
          mappings: Array(mappings),
```

elsom25: Not _strictly_ speaking, but practically for what I think is happening here.

dev/lib/product_taxonomy/models/integration_version.rb
```diff
@@ -98,17 +129,22 @@ def generate_distribution(output_path:, direction:)
...
     #
     # @param direction [Symbol] The direction of the distribution file to generate (:from_shopify or :to_shopify).
-    # @return [Hash]
+    # @return [Hash, Array<Hash>]
     def to_json(direction:)
+      if direction == :both
+        return [to_json(direction: :from_shopify), to_json(direction: :to_shopify)].compact
+      end
+
+      return @to_json[direction] if @to_json.key?(direction)
```
elsom25, line 138: I feel like it'd be better to just embrace the `if`'s of this method (channelling my inner sandi metz)

dev/lib/product_taxonomy/models/integration_version.rb
danielpgross: `Array(mappings)` actually doesn't behave as desired here, it coverts a hash input to an array (rather than wrapping it).

I still simplified this though by having the callers wrap the argument in an array 👍

tylerrowsell (approved):

danielpgross:
### Merge activity

* **Dec 5, 4:31 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Shopify/product-taxonomy/480).

PR #520, opened 2025-01-07
Ensure CLA check gets skipped for Argo

danielpgross:
The automatic CLA check isn't working as desired in some cases -- the check is being enforced even though Argo is the PR author. This is happening because the check currently checks if `github.actor` matches the Argo bot username, but `github.actor` is actually _the user who last caused CI to be triggered_, whether by commenting on the PR, pushing changes, etc.

[Here's a good example](https://github.com/Shopify/product-taxonomy/actions/runs/12656103875?pr=510) showing the check being enforced for an Argo PR because the triggering user was someone else.

The fix here is to directly check the PR author, instead of the event actor. This is done slightly differently depending on the event type:
* For a `pull_request_target` event: `github.event.pull_request.user.login`
* For an `issue_comment` event: `github.event.issue.user.login`

https://github.com/orgs/community/discussions/25502 provides more context.

### DIFF
--- a/.github/workflows/contributor_license_agreement.yml
+++ b/.github/workflows/contributor_license_agreement.yml
@@ -13,7 +13,12 @@ jobs:
       actions: write
       pull-requests: write
     if: |
-      github.actor != 'argo-translation-integration[bot]'
+      (
+        # The PR author is stored differently depending on the event type, see
+        # https://github.com/orgs/community/discussions/25502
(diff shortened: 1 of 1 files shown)


elsom25 (approved):
The joys of github workflows. Thanks for figuring this one out

danielpgross:
### Merge activity

* **Jan 7, 3:23 PM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Shopify/product-taxonomy/520).

PR #523, opened 2025-01-08
Trigger missing localizations for new locales

ricardotejedorsanz:
Dummy change to localization files to trigger a translation request for new locales.


Carlotta-Stefanelli (approved):
That's perfect. Thanks

PR #528, opened 2025-01-16
Generate dist files for new translations

ricardotejedorsanz:
Opening a separate PR to generate dist files that were causing argo test to fail in https://github.com/Shopify/product-taxonomy/pull/524 + quoting "ON" in polish to avoid script parsing error.


danielpgross (commented):
Looks like the behaviour for sorting `Other` isn't working correctly -- I looked at `fr` ("Autre") and see that it moved from the end of the file up next to others starting with "A".

I'll look into fixing this now and will post here once the fix is merged, so we can generate it again.

danielpgross:
Fix is in https://github.com/Shopify/product-taxonomy/pull/529

danielpgross (approved):
Looks great now 🚀

Let's be sure to squash the commits into a single one either before or as part of the merge into `main`, since the first commit has a huge diff that we don't need to keep.

PR #529, opened 2025-01-16
Fix `dist` sort order for localized `Value`s

danielpgross:
In #528, we discovered that new localized `dist` output for `Value`s had a different sort order than existing files.

The reason turned out to be that the old implementation sorted localized values according to their English names, while the new one sorted according to the localized names.

This PR changes the sort logic to match the previous behaviour: sorting by the English name, regardless of the localization. It also adds tests to cover this case.

I tested generating `dist` files for values with the `fr` locale and the changes were as expected.

### DIFF
--- a/dev/lib/product_taxonomy/models/value.rb
+++ b/dev/lib/product_taxonomy/models/value.rb
@@ -37,7 +37,7 @@ module ProductTaxonomy
       def to_json(version:, locale: "en")
         {
           "version" => version,
-          "values" => all_values_sorted(locale:).map { _1.to_json(locale:) },
+          "values" => all_values_sorted.map { _1.to_json(locale:) },
         }
       end
 
@@ -54,7 +54,7 @@ module ProductTaxonomy
         HEADER
         [
           header,
-          *all_values_sorted(locale:).map { _1.to_txt(padding:, locale:) },
+          *all_values_sorted.map { _1.to_txt(padding:, locale:) },
         ].join("\n")
       end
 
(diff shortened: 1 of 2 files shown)


danielpgross:
* **#529** <a href="https://app.graphite.dev/github/pr/Shopify/product-taxonomy/529?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/product-taxonomy/529?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

dev/lib/product_taxonomy/models/value.rb
```diff
@@ -86,8 +86,14 @@ def longest_gid_length
...
       end
 
-      def all_values_sorted(locale: "en")
-        all.sort_by { |value| [value.name(locale:) == "Other" ? 1 : 0, value.name(locale:), value.id] }
+      def all_values_sorted
+        all.sort_by do |value|
+          [
+            value.name(locale: "en").downcase == "other" ? 1 : 0,
+            value.name(locale: "en"),
+            value.id,
```
elsom25, line 94: I'd be okay with sorting by locale — but forcing `other` to the end?
```suggestion
            value.name(locale: "en").downcase == "other" ? 1 : 0,
            value.name(locale:),
            value.id,
```

elsom25: I guess it comes down to if we prefer sort-partity between languages or proper alphabetical 🤔

danielpgross: >I'd be okay with sorting by locale — but forcing other to the end?

That works, with the caveat that we'll get a very big diff the next time we generate dist due to the order changing.

>I guess it comes down to if we prefer sort-partity between languages or proper alphabetical 🤔

Right. One advantage of the old behaviour is that the ID ordering is stable across all locales.

elsom25: We can do the narrow one first — and decide in slack what ordering preference _we want_?

danielpgross: Agreed, good plan

elsom25 (approved):

danielpgross:
### Merge activity

* **Jan 16, 6:02 PM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Shopify/product-taxonomy/529).

PR #537, opened 2025-01-24
[HG][SG] Category name fixes

ricardotejedorsanz:
Creating this issue as a placeholder to fix two typos in category names:
- `gid://shopify/TaxonomyCategory/sg-1-7-10-2` `Sporting Goods > Athletics > Coaching & Officiating > Sport & Safety Whistles > Ginger Grip Whistles` should be **Finger Grip Whistles**
- `gid://shopify/TaxonomyCategory/hg-11-6-2-7-32` `Home & Garden > Kitchen & Dining > Kitchen Appliance Accessories > Coffee Maker & Espresso Machine Accessories > Coffee Maker & Espresso Machine Replacement Parts > Rubber Foots` should be **Rubber Feet**

Ideally, we shouldn't have to archive and create new categories, since there's no change in what they represent. We need to consider how this might impact bulk imports/exports as both the name and gid are supported, but aim to reduce toil to fix typos or style issues. cc @danielpgross


tylerrowsell:
According to big query there are 26 shops that use these categories (about 50 products). 

I would say these are okay because the likelihood of impact is very low. 
https://docs.google.com/spreadsheets/d/1BpnSWKkO1kzHCwL7tLRw8QGnVdYh1nSMYnKKoNLTqVo/edit?usp=sharing

ricardotejedorsanz:
> I also ran all of our nodes through ChatGPT to see if it would catch any other errors... haven't had a chance to review but here is all the outputs.

cc @ornanatan --we can review this together :pray:

ricardotejedorsanz:
In that case, given the minimal impact, I suggest we merge to fix those two typos in the current release while we work on a more systemic approach.

PR #542, opened 2025-01-28
Add `LocalizationsValidator` using validation in `Localized` mixin

danielpgross:
### TL;DR

Migrated `LocalizationsValidator` module from old tool version, reimplementing most of it for much better performance. 

On my machine, the new implementation takes ~150ms compared to ~22s for the old version. The speedup is thanks to using the already loaded localizations rather than loading them all again.

### What changed?

- Added a validation method to the `Localized` mixin
- Added a new `LocalizationsValidator` module to validate translations across categories, attributes, and values
- Integrated validation into the `generate_dist` command

[The old implementation](https://github.com/Shopify/product-taxonomy/blob/669c899e802daaff0127b3896641f63347e1edbc/app/commands/localizations_validator.rb) checked for 2 cases:

* Localization key missing
* Localization key present, but `name` key missing for it

These are effectively the same check, since the `name` key being present requires the localization key itself to be present. In the new implementation, they are combined into a single check.

### DIFF
--- a/dev/lib/product_taxonomy/models/mixins/localized.rb
+++ b/dev/lib/product_taxonomy/models/mixins/localized.rb
@@ -5,11 +5,29 @@ module ProductTaxonomy
     def localizations
       return @localizations if @localizations
 
-      humanized_name = name.split("::").last.humanize.downcase.pluralize
-      localization_path = File.join(ProductTaxonomy.data_path, "localizations", humanized_name, "*.yml")
+      localization_path = File.expand_path(
+        "localizations/#{localizations_humanized_model_name}/*.yml",
+        ProductTaxonomy.data_path,
+      )
       @localizations = Dir.glob(localization_path).each_with_object({}) do |file, localizations|
         locale = File.basename(file, ".yml")
-        localizations[locale] = YAML.safe_load_file(file).dig(locale, humanized_name)
+        localizations[locale] = YAML.safe_load_file(file).dig(locale, localizations_humanized_model_name)
+      end
+    end
+
+    # Validate that all localizations are present for the given locales. If no locales are provided, all locales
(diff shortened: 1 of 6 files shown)


dev/lib/product_taxonomy/models/mixins/localized.rb
```diff
@@ -27,6 +45,7 @@ def localizations
     #   name(locale: "en")
     #   description(locale: "en")
     def localized_attr_reader(*attrs, keyed_by: :friendly_id)
+      @localizations_keyed_by = keyed_by
```
elsom25, line 48: Seems risky to me to have an instance var, since this method can be called multiple times (using a different `keyed_by` for each batch of attrs)

danielpgross: We're using it up on L23 to fetch whatever the localizations file is keyed by. I don't think there's any other way to fetch that variable than storing it in a class instance var like this.

You're right that `localized_attr_reader` can _technically_ be called multiple times with different `keyed_by` arguments. But it was never intended that way -- what would it mean for a single localizations file to be keyed by multiple fields (e.g. ID and handle)? The intention was for it to be called a single time with all the localized fields and the `keyed_by` argument.

The right solution is probably to add a check to raise an error if `@localizations_keyed_by` is already set. Thoughts?

elsom25: Yeah that works — I do wish we had a pathway without the state, but good for now

danielpgross:
### Merge activity

* **Jan 29, 5:47 PM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Shopify/product-taxonomy/542).

PR #579, opened 2025-02-26
Migrate `add_value` command to new tool

danielpgross:
### TL;DR
Migrated command to add new attribute value to the taxonomy from deprecated tool version to new one.

### What changed?
- Migrated `AddValueCommand` to handle value creation logic
- Added `add_value` CLI command that accepts a name and attribute friendly ID
- Added `next_id` helper method to `Value` model for ID generation
- Added `add_value` method to `Attribute` model

### How to test?
1. Run the CLI command: `bin/product_taxonomy add_value "Blueish" "color"` to add a new value to the color attribute
2. Verify the value is created with correct ID, friendly_id, and handle
3. Confirm the value appears in the appropriate data files
4. Try adding a value to an extended attribute to verify it's blocked
5. Attempt to add a duplicate value to ensure it's prevented

Sample run:
```
❯ bin/product_taxonomy add_value "Blueish" "color"
Created value `Blueish` for attribute `Color`
Dumping attributes...
Updated `/Users/danielgross/product-taxonomy/data/attributes.yml`
Dumping values...
Updated `/Users/danielgross/product-taxonomy/data/values.yml`
Syncing EN localizations
Syncing values...
Wrote values localizations to /Users/danielgross/product-taxonomy/data/localizations/values/en.yml
Version: unstable
Generating sibling groups...
Generating category search index...
Generating attributes...
Generating mappings...
Generating attributes with categories...
Generating attribute with categories search index...
Completed in 2.78 seconds
```

PR #580, opened 2025-02-27
Improve `Indexed` behaviour with model subclasses

danielpgross:
### TL;DR

Improved the `Indexed` mixin with a new `create_validate_and_add!` convenience method and improved behaviour for subclassed models.

### Why?

While migrating `add_category`, `add_attribute` and `add_value` commands, I noticed a pattern:
1. Create a new model object
2. Validate it
3. Add it to the model index

It seemed like a good idea to create a method that wrapped these 3 operations together. Simple enough.

While doing that, though, I realized a quirk of the `Indexed` mixin with subclassed models. Take `Attribute` and `ExtendedAttribute` for example: they share an index, since we want to access them all together. But we still want `ExtendedAttribute` to have specialized behaviour, hence the dedicated subclass.

Before this PR, the indexing logic was being handled manually: a new `ExtendedAttribute` would need to be manually added to the `Attribute` index using `Attribute.add(extended_attribute)`. _**More confusing was that `ExtendedAttribute.all` would then return nothing**_, since it was operating on its own dedicated index that was never being populated.

A better behaviour would be for the index to be automatically shared between an `Indexed` class and its subclasses.

### How?

* `Indexed` uses the built-in Ruby `extended` hook to set a boolean class instance variable `@is_indexed`.
* `hashed_models` moves up the class inheritance chain until it finds a class with `@is_indexed == true`
* Since all other indexing operations (ex. `add`) operate on `hashed_models`, this ensures that the index is shared between the indexed class and its subclasses for _all_ operations.
* Now, calling `ExtendedAttributes.all` returns the same result as calling `Attributes.all`, reflecting that the index is shared.

### DIFF
--- a/dev/lib/product_taxonomy/models/mixins/indexed.rb
+++ b/dev/lib/product_taxonomy/models/mixins/indexed.rb
@@ -3,6 +3,14 @@
 module ProductTaxonomy
   # Mixin providing indexing for a model, using hashes to support fast uniqueness checks and lookups by field value.
   module Indexed
+    class << self
+      # Keep track of which model class used "extend Indexed" so that the model can be subclassed while still storing
+      # model objects in a shared index on the superclass.
+      def extended(indexed_model_class)
+        indexed_model_class.instance_variable_set(:@is_indexed, true)
+      end
+    end
+
     # Validator that checks for uniqueness of a field value across all models in a given index.
     class UniquenessValidator < ActiveModel::EachValidator
       def validate_each(record, attribute, value)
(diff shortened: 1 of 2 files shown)


danielpgross:
* **#581** <a href="https://app.graphite.dev/github/pr/Shopify/product-taxonomy/581?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#579** <a href="https://app.graphite.dev/github/pr/Shopify/product-taxonomy/579?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#580** <a href="https://app.graphite.dev/github/pr/Shopify/product-taxonomy/580?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/product-taxonomy/580?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

elsom25 (approved):

PR #581, opened 2025-02-27
Migrate `add_attribute` command to new tool

danielpgross:
### TL;DR

Migrated command to add new base and extended attributes to the taxonomy from deprecated tool version to new one.

### What changed?
- Migrated `AddAttributeCommand` to create new attributes with values, with support for creating both base attributes (with new values) and extended attributes (inheriting values from base attributes)
- Implemented `next_id` method in `Attribute` model to auto-generate IDs for new attributes

### How to test?

Add a new base attribute with values:
```
bin/product_taxonomy add_attribute "Alternative color" "Another type of color" --values "Blue,Yellow,Green"
```

Add an extended attribute by extending the base attribute:
```
bin/product_taxonomy add_attribute "Extended color" "An extended type of color" --base_attribute_friendly_id "alternative_color"
```

Sample run:
```
❯ bin/product_taxonomy add_attribute "Alternative color" "Another type of color" --values "Blue,Yellow,Green"
Created base attribute `Alternative color` with friendly_id=`alternative_color`
Dumping attributes...
Updated `/Users/danielgross/product-taxonomy/data/attributes.yml`
Dumping values...
Updated `/Users/danielgross/product-taxonomy/data/values.yml`
Syncing EN localizations
Syncing attributes...
Wrote attributes localizations to /Users/danielgross/product-taxonomy/data/localizations/attributes/en.yml
Syncing values...
Wrote values localizations to /Users/danielgross/product-taxonomy/data/localizations/values/en.yml
Version: unstable
Generating sibling groups...
Generating category search index...
Generating attributes...
Generating mappings...
Generating attributes with categories...
Generating attribute with categories search index...
Completed in 2.96 seconds
```


dev/lib/product_taxonomy/commands/add_attribute_command.rb
```diff
@@ -0,0 +1,87 @@
...
+        existing_value = Value.find_by(friendly_id: value_friendly_id)
+        next existing_value if existing_value
+
+        value = Value.create_validate_and_add!(
+          id: Value.next_id,
+          name: value_name,
+          friendly_id: value_friendly_id,
+          handle: IdentifierFormatter.format_handle(value_friendly_id),
+        )
+        value
```
elsom25, line 83: `.create_validate_and_add!` returns value right? So I think we only need:

```rb
Value.create_validate_and_add!(
  ...
)
```

dev/lib/product_taxonomy/models/attribute.rb
```diff
@@ -42,6 +42,11 @@ def sorted_base_attributes
         all.reject(&:extended?).sort_by(&:name)
       end
 
+      # Get the next ID for a newly created attribute.
+      #
+      # @return [Integer] The next ID.
+      def next_id = (all.map(&:id).max || 0) + 1
```
elsom25, line 48: Missed this in previous reviews, but I think we can be more efficient:

```rb
def next_id = (all.max_by(&:id).id || 0) + 1
```

I do wish we could have done something like `all.max(&:id)` but not supported syntax :shakes-fist:

danielpgross: Nice 👍 I updated this on `Attribute` and also `Value`

PR #590, opened 2025-03-05
Add `Indexed#find_by!` to reduce repetition

danielpgross:
### TL;DR
Added `find_by!` method to the `Indexed` mixin to standardize error handling for record lookups and reduce repetition across callsites.

### What changed?
- Added new `find_by!` method to the `Indexed` mixin that raises `NotFoundError` when records aren't found
- Replaced manual error handling in command classes with `find_by!` calls

### How to test?
1. Try to find a non-existent record using `find_by!`
2. Verify the error message includes the model name and search criteria
3. Run existing command tests to ensure they work with the new error handling
4. Verify that successful lookups still work as expected

### DIFF
--- a/dev/lib/product_taxonomy/models/mixins/indexed.rb
+++ b/dev/lib/product_taxonomy/models/mixins/indexed.rb
@@ -3,6 +3,8 @@
 module ProductTaxonomy
   # Mixin providing indexing for a model, using hashes to support fast uniqueness checks and lookups by field value.
   module Indexed
+    NotFoundError = Class.new(StandardError)
+
     class << self
       # Keep track of which model class used "extend Indexed" so that the model can be subclassed while still storing
       # model objects in a shared index on the superclass.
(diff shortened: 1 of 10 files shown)


dev/lib/product_taxonomy/models/mixins/indexed.rb
```diff
@@ -67,6 +69,20 @@ def find_by(**conditions)
       hashed_models[field][value]
     end
 
+    # Find a model by field value. Returns the first matching record or raises an error if not found. Only works for
+    # fields marked unique.
+    #
+    # @param conditions [Hash] Hash of field-value pairs to search by
+    # @return [Object] The matching model
+    def find_by!(**conditions)
+      record = find_by(**conditions)
```
elsom25, line 78: Should we use `...` here or do we want the conditions splat?

```rb
def find_by!(...)
  record = find_by(...)
```

danielpgross: Yep exactly, I tried the `...` first and then realized we need the params

PR #595, opened 2025-03-14
[HG][SG] Category name fixes - loc files

ricardotejedorsanz:
Fixes localization files following https://github.com/Shopify/product-taxonomy/pull/537

@danielpgross could a test have captured the inconsistency between `data/categories` and `data/localization/categories`?


ricardotejedorsanz:
> Fixes localization files following #537
> 
> @danielpgross could a test have captured the inconsistency between `data/categories` and `data/localization/categories`?

This would have followed the async process; probably fine as no change in meaning to the EN source (typo).

PR #596, opened 2025-03-14
[BI] Fix attribute name to Humidity control method

ricardotejedorsanz:
Fixes typo in attribute `Humidy control` as `Humidity control method`:

image

### DIFF
--- a/data/attributes.yml
+++ b/data/attributes.yml
@@ -30513,7 +30513,7 @@ base_attributes:
   - egg_turning__semi_automatic
   - egg_turning__other
 - id: 3297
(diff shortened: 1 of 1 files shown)


data/attributes.yml
```diff
@@ -30515,13 +30515,13 @@ base_attributes:
 - id: 3297
   name: Humidy control
```
danielpgross, line 30516: We also want to change this to `Humidity control method`, right? I think that's what's causing the integration tests to fail.

danielpgross (approved):

PR #600, opened 2025-03-19
[SG] Adding Curling category

ornanatan:
Curling is a sport in which players slide stones on a sheet of ice toward a target area that is segmented into four concentric circles.
The Equipment used for Curling is Curling brooms, stones (rocks), curling shoes.


than:
@ornanatan you are my hero 🥌

PR #622, opened 2025-04-03
Fix pl values

ornanatan:
Change Polish value from Błąd to Jesień


ricardotejedorsanz:
This test was failing because it expected spacing changes to the `dist` files:
image

I've fixed it here so we should be good to merge now: https://github.com/Shopify/product-taxonomy/pull/623

ricardotejedorsanz (approved):

PR #653, opened 2025-05-14
Only use one section for legacy Shopify mappings

willzeng274:
### TL;DR

Reorganized channel mappings to separate Shopify (Legacy) mappings from other mappings, effectively only using one section for all Shopify related mappings.

Related to #646

### What changed?

- Split the mappings into two categories: Shopify and non-Shopify mappings
- Created a dedicated section for all Shopify (Legacy) mappings
- Removed the conditional logic that was previously adding "(Legacy)" to individual Shopify mappings
- Improved the layout by grouping all Shopify legacy mappings under a single collapsible section

### How to test?

1. Navigate to a page that displays channel mappings
2. Verify that non-Shopify mappings appear first, each in their own collapsible section
3. Verify that all Shopify mappings are grouped under a single "Shopify (Legacy)" collapsible section
4. Expand both types of sections to ensure all mapping data is correctly displayed
5. Check that the breadcrumb information and category IDs are properly shown in both sections

### Why make this change?

This change improves the organization and readability of channel mappings by grouping related mappings together. By consolidating all Shopify legacy mappings under a single section, it reduces visual clutter and makes it easier for users to find the specific mapping information they need. This separation also provides a clearer distinction between current and legacy mapping systems!

### DIFF
--- a/docs/_includes/mapping_rules.html
+++ b/docs/_includes/mapping_rules.html
@@ -1,0 +1,24 @@
+{% assign mapping = include.mapping %}
+
+{% assign input_taxonomy = mapping.input_taxonomy %}
+{% assign output_taxonomy = mapping.output_taxonomy %}
+{% assign output_title_and_version = output_taxonomy | split: "/" %}
+{% assign output_title = output_title_and_version[0] %}
+{% assign output_version = output_title_and_version[1] %}
+{% assign rules = mapping.rules %}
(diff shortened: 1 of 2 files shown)


danielpgross (commented):
The approach here looks good 👍

danielpgross (approved):

PR #661, opened 2025-05-23
Integrations directory cleanup and update taxonomy mappings

willzeng274:
- Addresses everything in the issue other than the CI improvement

### Summary of changes:
- Added a method to clear the Shopify integrations directory before generating distribution files
- Fixed version handling in `generate_docs_command` to properly handle unstable versions
- Enhanced the taxonomy version management in mapping files by updating references to `input_taxonomy` and `output_taxonomy` instead of `input_version` and `output_version`
- Added a new method `update_taxonomy_in_from_shopify_mappings!` to handle taxonomy version updates during release generation
- Updated tests to verify proper taxonomy version handling in mapping files
- Improved the mapping validation test to dynamically discover Shopify taxonomy versions from disk

Workflow:


dev/lib/product_taxonomy/commands/generate_dist_command.rb
```diff
@@ -39,6 +41,16 @@ def execute
 
     private
 
+    def clear_shopify_integrations_directory
```
danielpgross, line 44: I think this behaviour fits better in the `IntegrationVersion` class, since that's the one responsible for generating the integration files. `IntegrationVersion#generate_distributions` seems like a good place for it.

Another advantage of having it there is that we could reuse the logic for generating the integrations path and avoid multiple sources of truth for it.

dev/lib/product_taxonomy/commands/generate_docs_command.rb
```diff
@@ -14,6 +14,8 @@ def initialize(options)
       super
 
       @version = validate_and_sanitize_version!(options[:version]) || UNSTABLE
+      # Versions ending with -unstable should just be "unstable" in docs
+      @version = UNSTABLE if @version.end_with?("-unstable")
```
danielpgross, line 18: Rather than handling it here, let's update the callsite in `GenerateReleaseCommand` to just leave out the argument when it's calling with an unstable version.

dev/lib/product_taxonomy/commands/generate_release_command.rb
```diff
@@ -115,16 +118,62 @@ def get_commit_hash(ref)
...
+        content = File.read(file)
+        logger.debug("Original content of #{file} (for taxonomy update):\n#{content}")
+
+        content.gsub!(/input_taxonomy: shopify\/\d{4}-\d{2}(-unstable)?/, "input_taxonomy: #{new_taxonomy_value}")
+
+        logger.info("Writing updated taxonomy content to #{file}...")
         File.write(file, content)
+        logger.debug("Updated content of #{file} (for taxonomy update):\n#{File.read(file)}")
       end
     end
```
danielpgross, line 178: I don't think we need this new method -- we should be able to reuse the existing `update_integration_mappings` method.

dev/lib/product_taxonomy/commands/generate_release_command.rb
```diff
@@ -115,16 +118,62 @@ def get_commit_hash(ref)
...
+
+        # Update input_taxonomy to the stable version (e.g., shopify/YYYY-MM)
+        stable_taxonomy_version = @version.gsub(/-unstable$/, '')
+        content.gsub!(/input_taxonomy: shopify\/.*?-unstable/, "input_taxonomy: shopify/#{stable_taxonomy_version}")
+        
+        logger.info("Writing updated content to #{file}...")
+        File.write(file, content)
+        logger.debug("Updated content of #{file}:\n#{File.read(file)}")
+      end
+    end
```
danielpgross, line 150: Do you plan to remove the `logger.debug` statements before shipping this?

dev/lib/product_taxonomy/commands/generate_release_command.rb
```diff
@@ -8,6 +8,10 @@ def initialize(options)
       @version = validate_and_sanitize_version!(options[:current_version])
       @next_version = validate_and_sanitize_version!(options[:next_version])
 
+      if @version.end_with?("-unstable")
+        raise ArgumentError, "Version must not end with '-unstable' suffix"
+      end
```
danielpgross, line 13: Can we add a unit test for this new check?

dev/lib/product_taxonomy/commands/generate_release_command.rb
```diff
@@ -88,11 +92,14 @@ def move_to_next_version!
       logger.info("Updating VERSION file to #{@next_version}...")
       File.write(version_file_path, @next_version)
 
+      logger.info("Updating input_taxonomy in from_shopify.yml mappings to #{@next_version}...")
```
danielpgross: I think it's okay to simply log "Updating integration mappings..." like we do in `generate_release_version!`. They're doing roughly the same operation, so they should be consistent.

dev/lib/product_taxonomy/commands/generate_release_command.rb
```diff
@@ -107,23 +114,43 @@ def run_git_command(*args)
...
+
+      logger.info("Updating from_shopify.yml files -> input_taxonomy: #{final_shopify_taxonomy_string}")
       Dir.glob(File.expand_path("integrations/**/mappings/from_shopify.yml", ProductTaxonomy.data_path)).each do |file|
+        logger.info("Processing file for input_taxonomy update: #{file}")
         content = File.read(file)
-        content.gsub!(/input_version: .*?-unstable/, "input_version: #{@version}")
+        content.gsub!(%r{input_taxonomy: shopify/\d{4}-\d{2}(-unstable)?}, "input_taxonomy: #{final_shopify_taxonomy_string}")
         File.write(file, content)
       end
     end
```
danielpgross: Let's remove all the `logger.info` and `logger.debug` statements in here -- other methods don't log in this detail and they make the methods longer than necessary.

dev/lib/product_taxonomy/commands/generate_release_command.rb
```diff
@@ -107,23 +114,43 @@ def run_git_command(*args)
...
         File.write(file, content)
       end
 
+      logger.info("Updating from_shopify.yml files -> input_taxonomy: #{final_shopify_taxonomy_string}")
+      Dir.glob(File.expand_path("integrations/**/mappings/from_shopify.yml", ProductTaxonomy.data_path)).each do |file|
+        logger.info("Processing file for input_taxonomy update: #{file}")
+        content = File.read(file)
+        content.gsub!(%r{input_taxonomy: shopify/\d{4}-\d{2}(-unstable)?}, "input_taxonomy: #{final_shopify_taxonomy_string}")
+        File.write(file, content)
+      end
```
danielpgross: There's an exact copy of this code below in `update_integration_mappings_next_unstable`, can we extract it into a shared method?

dev/test/commands/generate_release_command_test.rb
```diff
@@ -137,8 +137,68 @@ class GenerateReleaseCommandTest < TestCase
...
+      assert_match "output_taxonomy: shopify/#{@version}", to_shopify_content
+      refute_match "output_taxonomy: shopify/2023-12-unstable", to_shopify_content
+
+      # After execute, from_shopify should have the next unstable version
+      from_shopify_content = File.read(File.expand_path("data/integrations/shopify/mappings/from_shopify.yml", @tmp_base_path))
+      assert_match "input_taxonomy: shopify/#{@next_version}", from_shopify_content
+      refute_match "input_taxonomy: shopify/2023-12-unstable", from_shopify_content
+    end
+
+    test "mapping files transformations pipeline during execute stages" do
```
danielpgross: This test name seems pretty opaque, how about:

```suggestion
    test "execute updates input and output taxonomies for integrations in two stages" do
```

dev/lib/product_taxonomy/models/integration_version.rb
```diff
@@ -26,6 +28,20 @@ def generate_all_distributions(output_path:, logger:, current_shopify_version: n
...
+      # @param logger [Logger] The logger to use for logging messages.
+      def clear_shopify_integrations_directory(output_path:, logger:)
+        shopify_integrations_path = File.join(integrations_output_path(output_path), "shopify")
+        if Dir.exist?(shopify_integrations_path)
+          logger.info("Clearing Shopify integrations directory: #{shopify_integrations_path}")
+          # Note: currently we only generate json and txt files in this directory
+          # If anything changes in the future, this needs to be updated
+          FileUtils.rm_rf(Dir.glob(File.join(shopify_integrations_path, "*")))
+        end
+      end
```
danielpgross: We need unit tests covering the changes in this file

dev/lib/product_taxonomy/models/integration_version.rb
```diff
@@ -26,6 +28,20 @@ def generate_all_distributions(output_path:, logger:, current_shopify_version: n
...
+      #
+      # @param output_path [String] The path to the output directory.
+      # @param logger [Logger] The logger to use for logging messages.
+      def clear_shopify_integrations_directory(output_path:, logger:)
+        shopify_integrations_path = File.join(integrations_output_path(output_path), "shopify")
+        if Dir.exist?(shopify_integrations_path)
+          logger.info("Clearing Shopify integrations directory: #{shopify_integrations_path}")
+          # Note: currently we only generate json and txt files in this directory
+          # If anything changes in the future, this needs to be updated
+          FileUtils.rm_rf(Dir.glob(File.join(shopify_integrations_path, "*")))
```
danielpgross: Are you sure this glob works? I would expect it to only return files directly in that directory, but not in deeper ones. In other words, it's not recursive. We need it to operate recursively so that the `google` and `shopify` subdirectories get emptied too.

willzeng274: I think it deletes everything recursively, but just in case I'll just rm rf on the path and then recreate it

dev/lib/product_taxonomy/models/integration_version.rb
```diff
@@ -26,6 +28,19 @@ def generate_all_distributions(output_path:, logger:, current_shopify_version: n
...
       end
 
+      # Clear the Shopify integrations directory before generating new files.
+      #
+      # @param output_path [String] The path to the output directory.
+      # @param logger [Logger] The logger to use for logging messages.
+      def clear_shopify_integrations_directory(output_path:, logger:)
+        shopify_integrations_path = File.join(integrations_output_path(output_path), "shopify")
+        if Dir.exist?(shopify_integrations_path)
+          logger.info("Clearing Shopify integrations directory: #{shopify_integrations_path}")
```
danielpgross: Similar to the other comment, let's remove this log entry. We don't need to log every minor operation.

danielpgross (approved):

willzeng274:
### Note: This does not address the CI workflow that is supposed to execute on both commits

willzeng274:
### Merge activity

* **Jun 5, 9:22 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Shopify/product-taxonomy/661).
* **Jun 5, 9:34 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/Shopify/product-taxonomy/661) rebased this pull request as part of a merge.
* **Jun 5, 9:46 PM UTC**: @willzeng274 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Shopify/product-taxonomy/661).

PR #673, opened 2025-06-09
Fix CI to run tests on both release commits

willzeng274:
# CI Flow Changes

## Background

The product taxonomy release process creates two commits:
1. A "Release version" commit that finalizes the release
2. A "Bump version" commit that prepares for the next development cycle

Previously, CI would only run on the final commit in a PR, which meant the release commit wasn't properly tested.

This caused CI issues because:

- CI would only test the latest commit (bump commit) by default
- Testing both commits required complex CI configuration hacks
- The CI workflow became difficult to maintain and understand

- Previous GitHub action solution: ~~Example of CI working on a test release commit: https://github.com/Shopify/product-taxonomy/pull/677 (closed after deleting branch)~~

## Summary

This PR modifies the `generate_release_command` to create two separate branches (and thus two separate PRs) for the release process, replacing the previous approach of having both commits in a single branch.

## Solution

The command now creates **two separate branches**:

1. `release-v{version}` - Contains the release commit
2. `bump-v{next_version}` - Contains the bump to next unstable version

This enables a **stacked PR workflow** where each commit gets its own PR and CI runs naturally on each.

### Code Changes

1. **`generate_release_command.rb`**:

   - Added `run_git_command("checkout", "-b", "bump-v#{@next_version}")` in `move_to_next_version!`
   - Updated `print_summary` with instructions for creating and merging stacked PRs
   - Updated `print_rollback_instructions` to handle both branches

2. **`.github/workflows/tests.yml`**:

   - Reverted to simple CI configuration (no complex commit detection)
   - CI runs normally on all PRs

3. **Tests**:
   - Updated to expect the new branch creation
   - Updated expected output messages

### New Workflow

When you run the release command, you'll see:

```
====== Release Summary ======
- Created branch: release-v2024-01
  - Commit (abc1234): Release version 2024-01
  - Tag: v2024-01
- Created branch: bump-v2024-02-unstable
  - Commit (def5678): Bump version to 2024-02-unstable

Next steps:
  1. If the changes look good, push both branches
     * Run `git push origin release-v2024-01`
     * Run `git push origin bump-v2024-02-unstable`
  2. Open a PR and follow the usual process to get approval and merge the release branch
     * https://github.com/Shopify/product-taxonomy/pull/new/release-v2024-01
     * Target: main
  3. Open a PR and follow the usual process to get approval and merge the bump branch
     * https://github.com/Shopify/product-taxonomy/pull/new/bump-v2024-02-unstable
     * Target: release-v2024-01 (NOT main)
     * This PR should be set to merge into release-v2024-01
  4. Merge PRs in order
     * First: Merge the release PR into main
     * Second: Update the bump PR to target main, then merge
  5. Once the PRs are merged, push the tag that was created
     * Run `git push origin v2024-01`
  6. Create a release on GitHub
     * https://github.com/Shopify/product-taxonomy/releases/new?tag=v2024-01
```

## Rollback Improvements

The rollback instructions now properly handle the multi-branch scenario:

- First switches to main branch with `git checkout main`
- Then resets to remote state with `git reset --hard origin/main`
- Provides branch deletion commands for both branches if they exist

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,14 +22,17 @@ jobs:
       - uses: ruby/setup-ruby@v1
         with:
           bundler-cache: true
-          working-directory: ./dev
-      - name: Run test
-        working-directory: ./dev
+          working-directory: dev
+      - uses: cue-lang/setup-cue@v1.0.0
+        with:
+          version: "v0.7.0"
+      - name: Run tests
+        working-directory: dev
         run: |
-          if [ "${{ matrix.test }}" == "test_unit" ]; then
-            bundle exec rake test:unit
-          elif [ "${{ matrix.test }}" == "test_integration" ]; then
+          if [ "${{ matrix.test }}" = "test_integration" ]; then
             bundle exec rake test:integration
+          elif [ "${{ matrix.test }}" = "test_unit" ]; then
+            bundle exec rake test:unit
           else
             bundle exec rake schema:vet
           fi


willzeng274:
* **#673** <a href="https://app.graphite.dev/github/pr/Shopify/product-taxonomy/673?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/product-taxonomy/673?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.github/workflows/tests.yml
```diff
@@ -5,20 +5,57 @@ on:
   push:
     branches:
       - main
+      - 'release-v*'
   pull_request:
     branches:
       - main
-    types: [opened, synchronize, reopened, target]
+    types: [opened, synchronize, reopened]
```
willzeng274, line 11: The reason why I removed "target" is because the linter was yelling at me saying it's not a valid type. Let me know if this removal as wrong

danielpgross (commented):
@willzeng274 does this solution block the PR from being merged if the release commit is failing CI? That's the main goal -- we want to make sure a bad release commit doesn't get merged by accident.

willzeng274:
Yes, I actually tested it with a closed PR to check if all CI runs. Basically what this solution does is that it starts 3 test tasks (how it originally works) for each commit (finds commits based on their commit message). It will still only show that CI ran on the last commit (couldn't find a single solution to address this on stackoverflow or through AI), but it creates 6 (3 * 2) tasks to test both of the commits and if any of them fails it cannot be merged

.github/workflows/tests.yml
willzeng274: The rest of the changes in this file are autoformatted

willzeng274:
Okay, right now I've implemented a solution like this:
- Creates 2 separate branches for each commit
- Instructs the user to make separate PRs, with instructions for the user to set the target branch for the "Bump version" PR to the branch for the release PR to create a "stack"

But there are some other solutions:
1. GitHub Actions but using it differently
  - We can create a new workflow entirely (new .yml file inside of .github/workflows), it will still execute on every PR merging into main but I can add `if: startsWith(github.head_ref, 'release-v')` so the action will show as "skipped" on all PRs except a PR with branch starting with "release-v". There will still be major disadvantages including clarity (executes double the amount of test actions on the last commit, but we can add clarity by adding a prefix to their names ? not tested yet). With caching checkouts (not explore yet) or if we don't value CI performance on release-PRs it could potentially work.
2. Push each commit individually
- We can edit the workflow to run tests on push event with branches in a pattern like `release-v*`, and push those 2 commits separately within the pipeline. The disadvantage is that we lose control of when we want to push.
3. Using Graphite
- If we know for sure we will be using graphite long term, then we can run graphite CLI commands (or use an API if they have one) to create a stacked PR automatically. This gives a bit of flexibility compared to the double PR solution (we can make a draft PR, etc) but it restricts us within the Graphite ecosystem.

danielpgross:
We want to keep the release process as simple as possible to reduce the likelihood of mistakes during the process.

I looked more into it and found that we can create a ruleset protecting tags by enforcing that CI needs to pass. This achieves the main goal of ensuring that the release itself is correct, even if it still allows a bad release commit to be merged to main.

PR #679, opened 2025-06-10
Update create_previous_version_mappings to only update existing files

willzeng274:
### TL;DR

My previous PR updated the `output_taxonomy` field of **EVERY** `to_shopify.yml` file inside of `data/integrations/shopify/**/mappings`. This is not desired and it should only be done for the google mappings.
Now we only modify the "latest" version folder inside of integrations if the mappings exist (output_taxonomy=something-unstable).

Note that this is purposely placed before `DumpIntegrationFullNamesComand` because it looks for the latest version folder, and having the full names dump before this operation would result in unexpected behaviour.

### What changed?

- Added a new `create_previous_version_mappings` method that finds the latest version directory and updates its `to_shopify.yml` file to point to the new version
- Modified the `update_mapping_files` method to skip Shopify version directories when updating mapping files
- Added comprehensive tests to verify the new functionality works correctly

### How to test
- Run generate release on a new branch after disabling the pull command and check_git_state to see if it modifies the files inside of `data/integrations/**/to_shopify.yml` correctly

### DIFF
--- a/dev/lib/product_taxonomy/commands/generate_release_command.rb
+++ b/dev/lib/product_taxonomy/commands/generate_release_command.rb
@@ -131,7 +131,11 @@ module ProductTaxonomy
     end
 
     def update_mapping_files(filename, field_name, new_value)
-      Dir.glob(File.expand_path("integrations/**/mappings/#{filename}", ProductTaxonomy.data_path)).each do |file|
+      files = Dir.glob(File.expand_path("integrations/**/mappings/#{filename}", ProductTaxonomy.data_path))
+
+      files = [files.sort.last] if filename == "to_shopify.yml"
+
+      files.each do |file|
         content = File.read(file)
         content.gsub!(%r{#{field_name}: shopify/\d{4}-\d{2}(-unstable)?}, "#{field_name}: #{new_value}")
         File.write(file, content)
--- a/dev/test/commands/generate_release_command_test.rb
+++ b/dev/test/commands/generate_release_command_test.rb
@@ -14,7 +14,10 @@ module ProductTaxonomy
 
       # Create test files and directories
       FileUtils.mkdir_p(File.expand_path("dist", @tmp_base_path))
-      FileUtils.mkdir_p(File.expand_path("data/integrations/shopify/mappings", @tmp_base_path))
+      FileUtils.mkdir_p(File.expand_path("data/integrations/shopify/2023-12/mappings", @tmp_base_path))
+      FileUtils.mkdir_p(File.expand_path("data/integrations/shopify/2022-01/mappings", @tmp_base_path))
+      FileUtils.mkdir_p(File.expand_path("data/integrations/shopify/2022-10/mappings", @tmp_base_path))
+      FileUtils.mkdir_p(File.expand_path("data/integrations/google/2023-12/mappings", @tmp_base_path))
 
       # Create VERSION file
       File.write(@version_file_path, "2023-12")
@@ -31,11 +34,19 @@ module ProductTaxonomy
 
       # Create test mapping files
       File.write(
-        File.expand_path("data/integrations/shopify/mappings/to_shopify.yml", @tmp_base_path),
+        File.expand_path("data/integrations/shopify/2023-12/mappings/to_shopify.yml", @tmp_base_path),
         "output_taxonomy: shopify/2023-12-unstable",
       )
       File.write(
-        File.expand_path("data/integrations/shopify/mappings/from_shopify.yml", @tmp_base_path),
+        File.expand_path("data/integrations/shopify/2022-01/mappings/to_shopify.yml", @tmp_base_path),
+        "input_taxonomy: shopify/2022-01-unstable",
+      )
+      File.write(
+        File.expand_path("data/integrations/shopify/2022-10/mappings/to_shopify.yml", @tmp_base_path),
+        "input_taxonomy: shopify/2022-07-unstable",
+      )
+      File.write(
+        File.expand_path("data/integrations/google/2023-12/mappings/from_shopify.yml", @tmp_base_path),
         "input_taxonomy: shopify/2023-12-unstable",
       )
 
@@ -140,8 +151,8 @@ module ProductTaxonomy
 
       command.execute
 
-      to_shopify_content = File.read(File.expand_path("data/integrations/shopify/mappings/to_shopify.yml", @tmp_base_path))
-      from_shopify_content = File.read(File.expand_path("data/integrations/shopify/mappings/from_shopify.yml", @tmp_base_path))
+      to_shopify_content = File.read(File.expand_path("data/integrations/shopify/2023-12/mappings/to_shopify.yml", @tmp_base_path))
+      from_shopify_content = File.read(File.expand_path("data/integrations/google/2023-12/mappings/from_shopify.yml", @tmp_base_path))
 
       assert_equal "output_taxonomy: shopify/#{@version}", to_shopify_content
       assert_equal "input_taxonomy: shopify/#{@next_version}", from_shopify_content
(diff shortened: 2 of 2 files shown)


willzeng274:
* **#679** <a href="https://app.graphite.dev/github/pr/Shopify/product-taxonomy/679?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/product-taxonomy/679?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

dev/lib/product_taxonomy/commands/generate_release_command.rb
```diff
@@ -122,6 +122,7 @@ def get_commit_hash(ref)
     end
 
     def update_integration_mappings_stable_release
+      create_previous_version_mappings
       update_mapping_files("to_shopify.yml", "output_taxonomy", "shopify/#{@version}")
```
danielpgross: If we're skipping `shopify` directories in `update_mapping_files`, this call will always do nothing, since only shopify mappings use `to_shopify.yml`.

I think there's a simpler/cleaner approach, I'll suggest it in a separate comment.

dev/lib/product_taxonomy/commands/generate_release_command.rb
```diff
@@ -130,8 +131,32 @@ def update_integration_mappings_next_unstable
...
+      # Only update if the file already exists
+      return unless File.exist?(to_shopify_file)
+
+      content = File.read(to_shopify_file)
+      content.gsub!(/output_taxonomy: shopify\/\d{4}-\d{2}(-unstable)?/, "output_taxonomy: shopify/#{@version}")
+      File.write(to_shopify_file, content)
+    end
+
     def update_mapping_files(filename, field_name, new_value)
       Dir.glob(File.expand_path("integrations/**/mappings/#{filename}", ProductTaxonomy.data_path)).each do |file|
```
danielpgross: Instead of the new logic in `create_previous_version_mappings`, you should be able to:
- store the result of the glob in a local variable
- if `filename == "to_shopify.yml"`, reject all glob entries except the latest version

dev/lib/product_taxonomy/commands/generate_release_command.rb
```diff
@@ -131,7 +131,19 @@ def update_integration_mappings_next_unstable
...
+
+      if filename == "to_shopify.yml"
+        shopify_files = files.select { |f| f.match?(%r{/integrations/shopify/\d{4}-\d{2}/}) }
+        if shopify_files.any?
+          latest_file = shopify_files.sort.last
+          files = files.reject { |f| f.match?(%r{/integrations/shopify/\d{4}-\d{2}/}) } + [latest_file]
+        end
+      else
+        files = files.reject { |f| f.match?(%r{/integrations/shopify/\d{4}-\d{2}/}) }
+      end
```
danielpgross: We can make this much simpler if we assume that only shopify integration files use `to_shopify.yml`, while all other integrations use `from_shopify.yml`. This is a valid assumption given how the integrations are defined today. In that case, we can just:
1. perform the glob
2. sort
3. take the last item and discard the rest

With this approach, there's no need for multiple `select`/`reject` calls on the array of filenames.

willzeng274: Yeah that makes sense. I was under the assumption that we would have both to_shopify.yml and from_shopify.yml for other mappings later on

dev/lib/product_taxonomy/commands/generate_release_command.rb
```diff
@@ -131,7 +131,14 @@ def update_integration_mappings_next_unstable
     end
 
     def update_mapping_files(filename, field_name, new_value)
-      Dir.glob(File.expand_path("integrations/**/mappings/#{filename}", ProductTaxonomy.data_path)).each do |file|
+      files = Dir.glob(File.expand_path("integrations/**/mappings/#{filename}", ProductTaxonomy.data_path))
+
+
+      if filename == "to_shopify.yml"
+        files = [files.sort_by { |f| f[%r{/shopify/(\d{4}-\d{2})/mappings/}, 1] }.last]
```
danielpgross: Any reason we need the special `sort_by` logic here? Can't we just use plain `sort`, as in `files.sort.last`?

willzeng274: I just tested a bunch of things and `files.sort.last` does work in our case, but the condition for lexicographic sorting to always consistently work for the file globs is to have everything before `/YYYY-MM/mappings` to be constant

danielpgross: Right. It's safe to assume that only shopify mappings use `to_shopify.yml`, meaning that everything before `/YYYY-MM/mappings` will be constant. So let's go with that.

willzeng274: Done 👍🏻

dev/test/commands/generate_release_command_test.rb
```diff
@@ -140,8 +151,8 @@ class GenerateReleaseCommandTest < TestCase
 
       command.execute
 
-      to_shopify_content = File.read(File.expand_path("data/integrations/shopify/mappings/to_shopify.yml", @tmp_base_path))
-      from_shopify_content = File.read(File.expand_path("data/integrations/shopify/mappings/from_shopify.yml", @tmp_base_path))
+      to_shopify_content = File.read(File.expand_path("data/integrations/shopify/2023-12/mappings/to_shopify.yml", @tmp_base_path))
+      from_shopify_content = File.read(File.expand_path("data/integrations/google/mappings/from_shopify.yml", @tmp_base_path))
```
danielpgross: Since we've added a version to the path on the previous line, let's do it here too

dev/test/commands/generate_release_command_test.rb
```diff
@@ -152,12 +163,12 @@ class GenerateReleaseCommandTest < TestCase
...
 
       # After execute, to_shopify should have the stable version
-      to_shopify_content = File.read(File.expand_path("data/integrations/shopify/mappings/to_shopify.yml", @tmp_base_path))
+      to_shopify_content = File.read(File.expand_path("data/integrations/shopify/2023-12/mappings/to_shopify.yml", @tmp_base_path))
       assert_match "output_taxonomy: shopify/#{@version}", to_shopify_content
       refute_match "output_taxonomy: shopify/2023-12-unstable", to_shopify_content
 
       # After execute, from_shopify should have the next unstable version
-      from_shopify_content = File.read(File.expand_path("data/integrations/shopify/mappings/from_shopify.yml", @tmp_base_path))
+      from_shopify_content = File.read(File.expand_path("data/integrations/google/mappings/from_shopify.yml", @tmp_base_path))
```
danielpgross: Same comment about adding version

willzeng274: Okay yeah, since there isn't any sorting I didn't put a version. But I'll put one there for consistency

dev/test/commands/generate_release_command_test.rb
```diff
@@ -191,8 +202,8 @@ class GenerateReleaseCommandTest < TestCase
 
       command.send(:move_to_next_version!)
 
-      to_shopify_content_stage2 = File.read(File.expand_path("data/integrations/shopify/mappings/to_shopify.yml", @tmp_base_path))
-      from_shopify_content_stage2 = File.read(File.expand_path("data/integrations/shopify/mappings/from_shopify.yml", @tmp_base_path))
+      to_shopify_content_stage2 = File.read(File.expand_path("data/integrations/shopify/2023-12/mappings/to_shopify.yml", @tmp_base_path))
+      from_shopify_content_stage2 = File.read(File.expand_path("data/integrations/google/mappings/from_shopify.yml", @tmp_base_path))
```
danielpgross: Add version here too

dev/test/commands/generate_release_command_test.rb
```diff
@@ -176,8 +187,8 @@ class GenerateReleaseCommandTest < TestCase
       command.send(:check_git_state!)
       command.send(:generate_release_version!)
 
-      to_shopify_content_stage1 = File.read(File.expand_path("data/integrations/shopify/mappings/to_shopify.yml", @tmp_base_path))
-      from_shopify_content_stage1 = File.read(File.expand_path("data/integrations/shopify/mappings/from_shopify.yml", @tmp_base_path))
+      to_shopify_content_stage1 = File.read(File.expand_path("data/integrations/shopify/2023-12/mappings/to_shopify.yml", @tmp_base_path))
+      from_shopify_content_stage1 = File.read(File.expand_path("data/integrations/google/mappings/from_shopify.yml", @tmp_base_path))
```
danielpgross: Add version here too

danielpgross (approved):

willzeng274:
### Merge activity

* **Jun 13, 8:10 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Shopify/product-taxonomy/679).
* **Jun 13, 8:22 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/Shopify/product-taxonomy/679) rebased this pull request as part of a merge.
* **Jun 13, 8:34 PM UTC**: @willzeng274 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Shopify/product-taxonomy/679).

PR #705, opened 2025-07-30
Fix incorrect translation of "prints" in Spanish localization

rodrigoescandon:
Changed "huellas dactilares" (fingerprints) to "prints" on line 18198. The original term refers to printed artwork/images, not fingerprints. "Prints" is commonly used in Spanish art/design contexts and maintains  clarity for users browsing art categories. Other options could be "impresiones" or "láminas".


ricardotejedorsanz:
Thank you for your suggestion, @rodrigoescandon!
@aalexandreia Did our localization experts review this change in the last batch? 🙏

PR #726, opened 2025-09-05
Add compare categories command

ricardotejedorsanz:
## Add category comparison command

This PR adds a new CLI command to compare category changes between different versions of the taxonomy.

### What it does
- Compares `full_names.yml` from a specific version folder against `dist/en/categories.txt`
- Detects three types of changes:
  - **Additions**: New categories in the current taxonomy
  - **Archived**: Categories that were removed/archived
  - **Renames**: Categories with the same ID but different names
- Outputs results to a timestamped CSV file in the `exports/` folder

### Usage
```bash
# Basic usage
bundle exec ruby -I lib bin/product_taxonomy compare_categories 2025-03

# With custom output directory
bundle exec ruby -I lib bin/product_taxonomy compare_categories 2024-10 --output-dir exports

# Show help
bundle exec ruby -I lib bin/product_taxonomy help compare_categories
```

### Files changed
- Added `dev/lib/product_taxonomy/commands/compare_categories_command.rb`
- Updated CLI to include the new command
- Added `exports/` folder to `.gitignore`

### Example output
The command generates CSV files with columns: `type`, `id`, `old_name`, `new_name`

image


dev/lib/product_taxonomy/cli.rb
```diff
@@ -95,5 +95,11 @@ def add_attributes_to_categories(attribute_friendly_ids, category_ids)
     def add_value(name, attribute_friendly_id)
       AddValueCommand.new(options.merge(name:, attribute_friendly_id:)).run
     end
+
+    desc "compare_categories VERSION_FOLDER", "Compare category changes between full_names.yml and categories.txt"
+    option :output_dir, type: :string, default: "exports", desc: "Output directory for CSV file"
```
danielpgross, line 100: It would be more flexible and idiomatic to accept the full output path of the CSV file (filename included), not just output dir

dev/lib/product_taxonomy/cli.rb
```diff
@@ -95,5 +95,11 @@ def add_attributes_to_categories(attribute_friendly_ids, category_ids)
     def add_value(name, attribute_friendly_id)
       AddValueCommand.new(options.merge(name:, attribute_friendly_id:)).run
     end
+
+    desc "compare_categories VERSION_FOLDER", "Compare category changes between full_names.yml and categories.txt"
```
danielpgross, line 99: I think we can make the description clearer, maybe something like this:

```suggestion
    desc "compare_categories VERSION_FOLDER", "Generate a CSV report showing category changes between the current taxonomy and Shopify taxonomy VERSION"
```

dev/lib/product_taxonomy/cli.rb
```diff
@@ -95,5 +95,11 @@ def add_attributes_to_categories(attribute_friendly_ids, category_ids)
...
+    option :output_dir, type: :string, default: "exports", desc: "Output directory for CSV file"
+    def compare_categories(version_folder)
```
danielpgross, line 101: I find `version_folder` confusing here, isn't it just the version?

```suggestion
    def compare_categories(version)
```

.gitignore
```diff
@@ -3,3 +3,6 @@
 /docs/_data/unstable
 /docs/.jekyll-metadata
 node_modules
+
+# Ignore exports folder
+/exports
```
danielpgross, line 8: We don't need to add an `exports` folder, the user can choose to export the CSV to wherever they want. Let's remove this.

danielpgross (commented):
Approach looks good overall. What stood out:
- this approach can't recognize relocations, it will report them as archive + add new category. Maybe that's okay, but good to be aware of.
- Did you consider using mappings for this? They were built to capture basically this same information and are better because they capture relocations

We would also need unit tests added before merging

PR #736, opened 2025-09-25
Regenerating dist after localization changes

katbailey:
The dist files hadn't been regenerated since the latest localization changes.
This PR is the result of running `bin/product_taxonomy dist --locales=all`.


danielpgross (approved):
Reviewed the diff locally, LGTM 👍 

I also did some digging to understand what happened leading up to this:
- The new translations were added in https://github.com/Shopify/product-taxonomy/pull/725
- It's normal practice to ship a separate PR to regenerate `dist` after adding localizations, see ex. https://github.com/Shopify/product-taxonomy/pull/528. @ricardotejedorsanz it looks like we just missed this step this time.

For the future, we could consider:
- Expanding CI checks to include all locales (not just `en`)
- Making it standard practice to regenerate `dist` as part of the translations PR itself rather than a follow up. (In fact, the PR would be blocked from merging until `dist` was updated)

Issue #753, opened 2025-10-05
[Taxonomy Tree]:Ριχτάρια


ouuzi:
### Vertical

Home & Garden

### Change Type

New category or attribute

### Issue

There is missing category for Sofa Throws. also in greek which is "Ριχτάρια"

### Description

I didn't find a suitable category or fitting attributes for my product,

ornanatan:
Thanks for reaching out and for the feedback.
Chair & Sofa Throws will be available in our next taxonomy release.
Adress in #757

PR #757, opened 2025-10-05
[H&G] Create category for chair & sofa throws [#753]

ornanatan:
Chair & Sofa Throws include decorative throws used as accent pieces on furniture. These are primarily chosen for style and design rather than warmth
image


ricardotejedorsanz (approved):
Good addition, thanks!

PR #776, opened 2025-10-28
[HB] Parent attribute inheritance for category expansion from taxonomy evolution

ricardotejedorsanz:
Parent attribute inheritance for new categories introduced from taxonomy evolution.

Attributes assigned to the parent categories are automatically inherited by new leaves for consistency.

**Note:** Net new attribute additions and other attribute connections will be included in a separate step.


data/categories/hb_health_beauty.yml
```diff
@@ -2194,11 +2811,15 @@
 - id: hb-2-3-6
   name: Earring Holders
   children: []
-  attributes: []
+  attributes:
```
ornanatan, line 2827: I would add material attribute as well

data/categories/hb_health_beauty.yml
```diff
@@ -2168,19 +2777,27 @@
...
+  - color
+  - pattern
 - id: hb-2-3-3-3
   name: Necklace Trays
   children: []
-  attributes: []
+  attributes:
```
ornanatan, line 2799: I would add material attribute as well

data/categories/hb_health_beauty.yml
```diff
@@ -2168,19 +2777,27 @@
...
+  - color
+  - pattern
 - id: hb-2-3-3-4
   name: Ring Trays
   children: []
-  attributes: []
+  attributes:
```
ornanatan, line 2806: I would add material attribute as well

data/categories/hb_health_beauty.yml
```diff
@@ -2194,11 +2811,15 @@
...
+  - color
+  - pattern
 - id: hb-2-3-7
   name: Jewelry Rolls
   children: []
-  attributes: []
+  attributes:
```
ornanatan, line 2834: I think that material is missing

ricardotejedorsanz:
I added `material` to those categories, can you please take another look @ornanatan? 🙏

ornanatan (approved):
Look great!

PR #779, opened 2025-10-28
[EL] New attribute connections from taxonomy evolution with inheritance

ricardotejedorsanz:
Attribute connections introduced from taxonomy evolution.
Attributes assigned to parent categories are automatically inherited by new leaves for consistency.

**Note:** Net new attribute additions will be included in a separate step.

### DIFF
--- a/data/categories/el_electronics.yml
+++ b/data/categories/el_electronics.yml
@@ -60,7 +60,6 @@
   - pattern
   - payment_mechanism_type
   - power_source
-  - sports_logo
 - id: el-1-2
   name: Pinball Machine Accessories
   children:


data/categories/el_electronics.yml
```diff
@@ -43,16 +43,21 @@
...
   attributes:
+  - ball_size
   - cabinet_type
   - color
+  - display_technology
   - frame_material
+  - height_adjustment
   - pattern
   - power_source
+  - sports_logo
```
ornanatan: I am not sure if this attribute is really necessary for this category

ricardotejedorsanz: Good catch! I've removed it.

ornanatan: Thanks!

ornanatan (approved):
LGTM

PR #810, opened 2025-11-20
[SG] Fix category name in Ice Climbing Tools area

ricardotejedorsanz:
Fix bug in naming of category `sg-4-3-13-3`: should be `Ice Climbing Screws` instead of `Ice Climbing Screwdrivers`

image


ornanatan (approved):
Perfect, LGTM

PR #811, opened 2025-11-21
Add ETL gem for graph construction

KaiKitJeffreyChan:
Resolves: https://github.com/shop/issues-custom-data-and-assets/issues/1625
Resolves: https://github.com/shop/issues-custom-data-and-assets/issues/1626

### TL;DR

[Vault project](https://vault.shopify.io/gsd/projects/48228-Products-Improving-the-taxonomy-pipeline)

Creates a Ruby gem for the Product Taxonomy ETL utilities and adds a GitHub workflow to publish it to RubyGems.

Screenshot 2025-11-24 at 4 23 20 PM

### Overview

Publish public repo ETL code as a small, code-only public RubyGems package

- Gem name: `shopify_product_taxonomy`
- Ship only dev/lib/** (no data/, dist/, or docs)
- Decouple versions from taxonomy data: cut releases only on meaningful ETL changes via `bump_version.yml`​github action workflow
- GitHub Action in public repo builds and pushes to RubyGems using trusted publishing
- Publishing a new gem version is simple: run `bump_verison.yml`​ with version number as input

### What changed?

- Added `shopify_product_taxonomy.gemspec`​ to package the ETL utilities as a Ruby gem
- Created two github action workflows 
    - `build_gem.yml`​ which verifies that the gem can be build successfully
    - `publish_gem.yml`​publishes gem to rubygems using trusted publishing when `version.rb` is changed
- Refactored taxonomy loading logic into a new `ProductTaxonomy::Loader` class that provides methods to construct the in-memory relations graph
- Made the data path configurable by adding a setter and using a default value
- Removed the `debug` and `thor` dependencies and `cli` commands from the main module

### How to test?

1. Verify the gem builds locally with `gem build shopify_product_taxonomy.gemspec`
2. Test the loader with custom data paths:
3. Create a PR that modifies the gemspec to test the workflow in verification mode

### Why make this change?

This change enables the Product Taxonomy ETL utilities to be distributed as a standalone Ruby gem, making it easier for other projects to consume and use the taxonomy data processing capabilities without needing to include the entire repository. The refactoring of the loader logic improves code organization and allows for more flexible configuration of data paths.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,309 @@
+## 2025-11
+
+#### 📚 Taxonomy Tree
+Expanded taxonomy with 1,000+ new categories and 600+ new attributes, plus extensive enhancements to existing categories.
+
+##### Apparel & Accessories
+
+- Adds the attributes `Gemstone type`, `Leather type` and `Robe style`.
+
+##### Arts & Entertainment
+
+- Adds `Kilns` (`Ceramic & Pottery Kilns`, `Enameling Kilns`, `Glass Kilns` and `Multipurpose Kilns`).
+- In the Percussion area, adds `Crash/Ride Cymbals`.
+- Adds the attributes `Kiln features`, `Kiln loading style`, `Material origin`, `Product classification` and `Product use`.
+
+##### Electronics
+
+- Adds `Joystick Parts & Accessories` (`Joystick Handles`, `Joystick Levers`, `Joystick Springs`, `Joystick Wiring Harnesses` and `Restrictor Gates`).
+- Adds `Fish Finder Accessories` (`Fish Finder Mounts` and `Fish Finder Transducers`).
+- Adds `DJ & Specialty Audio Accessories` (`DJ Flight Cases & Bags` and `DJ Skins & Covers`).
+- Adds `Coaxial Splitters & Switches` (`Coaxial Splitters` and `Coaxial Switches`).
+- Adds `3.5 mm Splitters & Switches` (`3.5 mm Splitters` and `3.5 mm Switches`).
+- Adds `Component Video Splitters & Switches` (`Component Video Splitters` and `Component Video Switches`).
+- Adds `DisplayPort Splitters & Switches` (`DisplayPort Splitters` and `DisplayPort Switches`).
+- Adds `RCA Splitters & Switches` (`RCA Splitters` and `RCA Switches`).
+- Adds `SDI Splitters & Switches` (`SDI Splitters` and `SDI Switches`).
+- Adds `Toslink Splitters & Switches` (`Toslink Splitters` and `Toslink Switches`).
+- Adds `Keyboard Keys` (`Artisan Keycaps` and `Keyboard Keycap Sets`).
+- In the Electronics area, adds `GPS Replacement Parts`.
+- In the Arcade Equipment area, adds 6 new categories including `Bingo Machines`, `Claw Crane Machines` and `Coin Pusher Machines`.
+- In the Pinball Machine Accessories area, adds 9 new categories including `Decals & Art Blades`, `Flipper Parts` and `Leg Levelers`.
+- In the Video Game Arcade Cabinet Accessories area, adds `Arcade Cabinet Artwork`, `Control Panel Protectors`, `Spinners`, `Trackballs` and `Wiring & Harnesses`.
+- In the Coin Mechanisms area, adds `Bill Acceptors & Validators`, `Coin Acceptors & Validators` and `Coin Doors`.
+- In the Control Panels area, adds `Control Panel Enclosure Kits` and `Control Panel Overlays`.
+- In the GPS Tracking Devices area, adds `Bluetooth Trackers` and `Vehicle GPS Trackers`.
+- In the Marine Electronics area, adds 8 new categories including `AIS Transceivers & Splitters`, `Marine AIS Receivers & Transceivers` and `Marine Amplifiers`.
+- In the Marine Chartplotters & GPS area, adds `Handheld Marine GPS Units`.
+- In the Marine Radar area, adds `Marine Radar Accessories`, `Marine Radar Units` and `Radar Reflectors`.
+- In the Marine Speakers area, adds `Marine Coaxial Speakers`, `Marine Subwoofers` and `Marine Tower Speakers`.
+- In the Bridges & Routers area, adds `Cellular Routers`, `Mesh WiFi Systems` and `Wired Routers`.
+- In the Repeaters & Transceivers area, adds `Ethernet & PoE Extenders`, `Fiber Optic Transceiver Modules` and `Wi-Fi Range Extenders`.
+- In the Hubs & Switches area, adds `Ethernet Hubs` and `Media Converters`.
+- In the Modular Switches area, adds `Network Switch Expansion Modules`, `Switch Fan Modules`, `Switch Power Supplies` and `Switch Stacking Kits`.
+- In the Modems area, adds `Cable Modems`, `Cellular Modems` and `Satellite Modems`.
+- In the Network Cards & Adapters area, adds `Bluetooth Network Adapters`, `Ethernet Cards & Adapters`, `Powerline Network Adapters` and `Wi-Fi Network Cards & Adapters`.
+- In the Power Over Ethernet Adapters area, adds `Power Over Ethernet Converters`, `Power Over Ethernet Extenders`, `Power Over Ethernet Injectors` and `Power Over Ethernet Splitters`.
+- In the Print, Copy, Scan & Fax area, adds `3D Printer Replacement Parts`.
+- In the 3D Printer Accessories area, adds `3D Printer Adhesives`, `3D Printer Enclosures` and `Filament Dryers`.
+- In the 3D Printer Cleaning Kits area, adds `3D Printer Cleaning Filaments`, `3D Printer Cleaning Fluids & Sprays` and `3D Printer Cleaning Tool Kits`.
+- In the Printer, Copier & Fax Machine Accessories area, adds `Automatic Document Feeders`, `Printer Cables & Power Cords`, `Printer Mounts & Brackets`, `Printer Platens` and `Rotary & Bed Attachments`.
+- In the Printer Consumables area, adds `Fuser & Roller Maintenance Kits`, `Maintenance Cartridges & Tanks`, `Printer Cleaning & Flushing Solutions`, `Printer Media & Film` and `Waste Toner Containers`.
+- In the Printer, Copier & Fax Machine Replacement Parts area, adds `Fuser Units`, `Paper Trays & Feeders`, `Printer Belts`, `Printer Replacement Cables` and `Printer Rollers`.
+- In the Printers, Copiers & Fax Machines area, adds `Copiers`, `Fax Machines`, `Inkjet Printers`, `Laser Printers` and `Multifunction Printers`.
+- In the Scanner Accessories area, adds `Film Scanning Accessories`, `Scanner Calibration Sheets` and `Scanner Roller Kits`.
+- In the Scanners area, adds `3D Scanners` and `Book Scanners`.
+- In the Sheetfed Scanners area, adds `Check Scanners`, `ID Card Scanners` and `Overhead Book Scanners`.
+- In the Video area, adds `Interactive Displays & Whiteboards` and `Video Wall Displays & Packages`.
+- In the Computer Monitors area, adds `Portable Monitors`.
+- In the Video Transmitters area, adds `AV Over IP Encoders`, `FPV Video Transmitters`, `HDBaseT Extenders` and `Wireless HDMI Extenders`.
+- In the Projectors area, adds `Film Projectors`.
+- In the Satellite & Cable TV area, adds `Satellite Dishes`, `Satellite Signal Meters & Finders` and `Satellite TV Kits`.
+- In the Video Accessories area, adds `Teleprompters` and `Video Magnifiers`.
+- In the Computer Monitor Accessories area, adds 6 new categories including `Color Calibration Charts`, `Monitor Arms & Mounts` and `Monitor Hoods & Sun Hoods`.
+- In the Projector Accessories area, adds `Projector Air Filters`, `Projector Cases & Bags` and `Projector Lenses`.
+- In the Television Parts & Accessories area, adds `TV Covers`, `TV Decorative Frames & Bezels`, `TV LED Backlight Strips & Bars` and `TV Replacement Stands & Legs`.
+- In the Video Editing Hardware & Production Equipment area, adds `Video Processors & Scalers`, `Video Switchers & Mixers` and `Video Wall Controllers`.
+- In the Video Players & Recorders area, adds `DVD & VCR Combo Players`, `DVD & VHS Combo Recorders`, `LaserDisc Players` and `Network Video Recorders`.
+- In the DVD & Blu-ray Players area, adds `Portable DVD Players`.
+- In the Video Game Console Accessories area, adds `Interactive Gaming Figures`, `Sim Racing Cockpits`, `Video Game Console Cables`, `Video Game Controllers` and `Virtual Reality Headset Accessories`.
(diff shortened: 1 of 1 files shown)


KaiKitJeffreyChan:
* **#811** <a href="https://app.graphite.com/github/pr/Shopify/product-taxonomy/811?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/product-taxonomy/811?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.github/workflows/publish_gem.yml
```diff
@@ -0,0 +1,36 @@
+name: Publish ETL Gem
+
+on:
+  push:
+    tags:
+      - 'etl-v*'
+  pull_request:
```
rafaelfranca: We should split the build verification and the publishing in two workflows. We can't trust the build verification in the same way we trust the publishing.

dev/shopify_product_taxonomy.gemspec
rafaelfranca, line 1: You probably should move the gemspec to inside the `dev` folder. It isn't following standard in Ruby libraries, live using `lib` as the `require_paths`

product_taxonomy-etl.gemspec
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+Gem::Specification.new do |spec|
+  spec.name          = "product_taxonomy-etl"
+  
+  # Extract version from git tag (e.g., etl-v1.2.0 -> 1.2.0)
+  # Falls back to "0.0.0" if no matching tag is found
+  spec.version       = `git describe --tags --match "etl-v*" 2>/dev/null`.strip.sub(/^etl-v/, '') || "0.0.0"
```
rafaelfranca: It is better if the gemspec don't depend on git to work.

rafaelfranca (changes requested):

dev/lib/product_taxonomy/loader.rb
```diff
@@ -0,0 +1,55 @@
...
+        categories_source_data = categories_glob.each_with_object([]) do |file, array|
+          array.concat(YAML.safe_load_file(file))
+        end
+        ProductTaxonomy::Category.load_from_source(categories_source_data)
+      end
+
+      def validate_taxonomy
+        # Run validations that can only be run after the taxonomy has been loaded.
+        ProductTaxonomy::Value.all.each { |model| model.validate!(:taxonomy_loaded) }
+      end
```
danielpgross: Somewhat subjective, but I don't like one-line methods -- they add needless indirection and additional code. I would just put the code inline in `load`.

dev/lib/product_taxonomy.rb
```diff
@@ -3,17 +3,21 @@
 require "thor"
 require "active_support/all"
 require "active_model"
-require "debug"
 
 module ProductTaxonomy
-  DATA_PATH = File.expand_path("../../data", __dir__)
-  private_constant :DATA_PATH
+  # Default to repo structure for backward compatibility
```
danielpgross: Comment is misleading: this is the main use-case, not backwards-compatibility. Let's remove.

dev/test/commands/command_test.rb
```diff
@@ -47,7 +47,7 @@ def execute
 
       command = TestCommand.new({})
       mock_value = mock("value")
-      ProductTaxonomy::Value.expects(:all).returns([mock_value])
+      ProductTaxonomy::Value.expects(:all).twice.returns([mock_value])
```
danielpgross: where was the second call added?

dev/lib/product_taxonomy/loader.rb
```diff
@@ -0,0 +1,55 @@
...
+          end
+
+          validate_taxonomy
+
+          {
+            categories: ProductTaxonomy::Category.all,
+            attributes: ProductTaxonomy::Attribute.all,
+            values: ProductTaxonomy::Value.all,
+            verticals: ProductTaxonomy::Category.verticals,
+          }
```
danielpgross: The return value is unused, this is a void method that populates the singleton classes with data. Let's remove.

```suggestion
```

dev/product_taxonomy-etl.gemspec
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+Gem::Specification.new do |spec|
+  spec.name          = "product_taxonomy-etl"
```
danielpgross: why the mix of underscore and hyphen?

dev/product_taxonomy-etl.gemspec
```diff
@@ -0,0 +1,34 @@
...
+  
+  # Extract version from git tag (e.g., etl-v1.2.0 -> 1.2.0)
+  # Falls back to "0.0.0" if no matching tag is found
+  spec.version       = `git describe --tags --match "etl-v*" 2>/dev/null`.strip.sub(/^etl-v/, '') || "0.0.0"
+  
+  spec.authors       = ["Shopify"]
+  spec.email         = ["taxonomy@shopify.com"]
```
danielpgross: does this email work?

product_taxonomy-etl.gemspec
```diff
@@ -0,0 +1,34 @@
...
+  
+  spec.authors       = ["Shopify"]
+  spec.email         = ["taxonomy@shopify.com"]
+
+  spec.summary       = "ETL utilities for Shopify Product Taxonomy"
```
danielpgross: We should expand the "ETL" acronym here and explain a bit more about what it does so that someone without context can understand.

product_taxonomy-etl.gemspec
```diff
@@ -0,0 +1,34 @@
...
+  spec.license       = "MIT"
+  spec.required_ruby_version = ">= 3.0"
+
+  spec.metadata = {
+    "bug_tracker_uri" => "https://github.com/Shopify/product-taxonomy/issues",
+    "source_code_uri" => "https://github.com/Shopify/product-taxonomy",
+    "allowed_push_host" => "https://rubygems.org",
+  }
+
+  spec.files         = Dir.glob("dev/lib/**/*")
```
danielpgross: should we exclude `commands` from here?

product_taxonomy-etl.gemspec
KaiKitJeffreyChan: I've removed the git dependency by introducing a standard version.rb file, which the publishing workflow automatically updates from the git tag during release.

dev/product_taxonomy-etl.gemspec
KaiKitJeffreyChan: underscore : Represents a multi-word name (e.g., active_record -> ActiveRecord).
hyphen: Represents an extension or a namespace structure (e.g., rspec-rails extends rspec).
gem naming conventions

dev/product_taxonomy-etl.gemspec
KaiKitJeffreyChan: I meant to ask this but I forgot, i put this for now but is there a better email I should put?

product_taxonomy-etl.gemspec
KaiKitJeffreyChan: Mhm yea makes sense, since the commands are internal CLI tools for maintaining the repo (generating docs, adding categories) and arent needed by external consumers who just want to load and process data.

dev/product_taxonomy-etl.gemspec
danielpgross: I had a look at [our RubyGems organizational profile](https://rubygems.org/profiles/shopify), which shows `gems@shopify.com` -- let's go with that.

dev/product_taxonomy-etl.gemspec
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+Gem::Specification.new do |spec|
+  spec.name          = "product_taxonomy-etl"
```
danielpgross: Thinking about this more...

First, I don't think we need `etl` in the name, since we're only publishing a single product taxonomy gem. Also, since this gem name will be part of the global gem namespace (i.e. gems outside Shopify), it might be helpful to have `shopify` in it.

Some ideas:
`shopify-product_taxonomy`
`product_taxonomy_graph`

dev/product_taxonomy-etl.gemspec
```diff
@@ -0,0 +1,35 @@
...
+
+  require_relative "lib/product_taxonomy/version"
+  spec.version       = ProductTaxonomy::VERSION
+  
+  spec.authors       = ["Shopify"]
+  spec.email         = ["taxonomy@shopify.com"]
+
+  spec.summary       = "Extract, transform, and load utilities for Shopify's Standard Product Taxonomy"
+  spec.description   = "A code-only gem providing Ruby classes and utilities to parse and process " \
+                       "Shopify's Standard Product Taxonomy data (categories, attributes, and values) " \
```
danielpgross: This also includes mappings and other future data like return reasons, so we don't need to enumerate them here.

```suggestion
                       "Shopify's Standard Product Taxonomy data " \
```

dev/product_taxonomy-etl.gemspec
```diff
@@ -0,0 +1,35 @@
...
+Gem::Specification.new do |spec|
+  spec.name          = "product_taxonomy-etl"
+
+  require_relative "lib/product_taxonomy/version"
+  spec.version       = ProductTaxonomy::VERSION
+  
+  spec.authors       = ["Shopify"]
+  spec.email         = ["taxonomy@shopify.com"]
+
+  spec.summary       = "Extract, transform, and load utilities for Shopify's Standard Product Taxonomy"
```
danielpgross:
```suggestion
  spec.summary       = "Load the complete Shopify Standard Product Taxonomy into memory as a tree"
```

.github/workflows/build_gem.yml
danielpgross, line 1: How about `build_gem.yml` instead?

dev/product_taxonomy-etl.gemspec
KaiKitJeffreyChan: Makes sense, using `shopify-product_taxonomy`

KaiKitJeffreyChan:
For discussion and review: Do we think `shopify_product_taxonomy` is a suitable name for the gem?

dev/lib/product_taxonomy/version.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+module ProductTaxonomy
+  VERSION = "0.0.0"
```
danielpgross: Let's start with `1.0.0` since the code is already feature-complete and stable

.github/workflows/bump_gem_version.yml
```diff
@@ -0,0 +1,47 @@
...
+      - name: Update version.rb
+        run: |
+          sed -i 's/VERSION = ".*"/VERSION = "${{ inputs.version }}"/' dev/lib/product_taxonomy/version.rb
+          echo "Updated version to:"
+          grep "VERSION =" dev/lib/product_taxonomy/version.rb
+
+      - name: Commit and push
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
```
danielpgross: Looks like this is current best practice ([reference](https://github.com/actions/checkout?tab=readme-ov-file#push-a-commit-using-the-built-in-token)):

```suggestion
          git config user.name "github-actions[bot]"
          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```

dev/lib/product_taxonomy.rb
```diff
@@ -3,17 +3,21 @@
 require "thor"
```
danielpgross: We can move the `thor` dependency into `command.rb`, since it's not needed when using the gem on its own.

dev/shopify_product_taxonomy.gemspec
```diff
@@ -0,0 +1,35 @@
...
+    "bug_tracker_uri" => "https://github.com/Shopify/product-taxonomy/issues",
+    "source_code_uri" => "https://github.com/Shopify/product-taxonomy",
+    "allowed_push_host" => "https://rubygems.org",
+  }
+
+  spec.files         = Dir.glob("lib/**/*").reject { |f| f.start_with?("lib/product_taxonomy/commands") }
+  spec.require_paths = ["lib"]
+
+  # Runtime dependencies from dev/Gemfile
+  spec.add_dependency "thor", "~> 1.4"
```
danielpgross: This can be removed after moving the `require` into `command.rb`

.github/workflows/build_gem.yml
```diff
@@ -0,0 +1,24 @@
...
+  build:
+    name: Verify Gem Build
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - uses: ruby/setup-ruby@d5126b9b3579e429dd52e51e68624dda2e05be25 # v1.267.0
+        with:
+          ruby-version: '3.2'
```
jenshenny: the Ruby version isn't needed (it'll take it from `.ruby-version`, same with the publish gem workflow)

```suggestion
        with:
          bundler-cache: true
```

dev/shopify_product_taxonomy.gemspec
```diff
@@ -0,0 +1,33 @@
...
+  spec.license       = "MIT"
+  spec.required_ruby_version = ">= 3.0"
+
+  spec.metadata = {
+    "bug_tracker_uri" => "https://github.com/Shopify/product-taxonomy/issues",
+    "source_code_uri" => "https://github.com/Shopify/product-taxonomy",
+    "allowed_push_host" => "https://rubygems.org",
+  }
+
+  spec.files         = Dir.glob("lib/**/*").reject { |f| f.start_with?("lib/product_taxonomy/commands") || f == "lib/product_taxonomy/cli.rb" }
```
rafaelfranca, line 27: Don't you need to include the data files in this gem? Otherwise they will not be able to be loaded.

KaiKitJeffreyChan: This gem intentionally excludes data files. It provides only the Ruby classes and utilities to load and manipulate taxonomy data - not the data itself.

.github/workflows/build_gem.yml
KaiKitJeffreyChan: CI fails with: `Error: input ruby-version needs to be specified if no .ruby-version or .tool-versions file exists`
I think CI fails without ruby-version because .ruby-version is in `dev/`, not the repo root. working-directory only affects bundler, not `.ruby-version` detection. Should I add a root-level `.ruby-version` instead?

dev/shopify_product_taxonomy.gemspec
danielpgross: The data files are excluded because:

- they're released on a different (more frequent) cadence from the loader code (gem)
- they would make the gem significantly bigger

danielpgross:
```
❯ du -sh ./data
160M    ./data
```

That's without any compression, but still

.github/workflows/build_gem.yml
jenshenny: I think it should work with setting the `working-directory` for that action https://github.com/ruby/setup-ruby?tab=readme-ov-file#working-directory

```
      - uses: ruby/setup-ruby@d5126b9b3579e429dd52e51e68624dda2e05be25 # v1.267.0
        working-directory: dev
        with:
          bundler-cache: true
 ```

File: dev/shopify_product_taxonomy.gemspec (hunk 19 above)
Comment (rafaelfranca) [reply dev/shopify_product_taxonomy.gemspec:27]: Is it documented in the gem how to get the data?

Comment (KaiKitJeffreyChan) [reply dev/shopify_product_taxonomy.gemspec:27]: Just updated the `spec.description` to specify where the data files are kept.

File: .github/workflows/publish_gem.yml [hunk 20]
```diff
@@ -0,0 +1,35 @@
+name: Publish Gem
+
+on:
+  workflow_call:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+    paths:
+      - 'dev/lib/product_taxonomy/version.rb'
```
rafaelfranca, line 10: You should only dispatch this event in the release publishing. Otherwise you might release without a reason.

Issue #808, opened 2025-11-19
[Taxonomy Tree]: Furniture (Sofa beds)


hudathudat:
### Vertical

Furniture

### Change Type

New category or attribute

### Issue

Add "sofa beds" as a level 2 category under sofas

### Description

Our hero product, sofa beds, do not fit into any current category. The sofa bed is a product that can convert between a sofa and a bed.
e.g. https://au.koala.com/products/koala-sofa-bed?variant=43934385701000

### Store Public URL

https://au.koala.com/products/koala-sofa-bed?variant=43934385701000

ornanatan:
Thank you for reaching out. 
We've reviewed your request and have added the Sofa Beds category to our taxonomy tree. You will see these changes reflected in our next taxonomy release.
You can see the detailes in ticket #812

PR #812, opened 2025-11-24
Add a category for sofa beds [#808]

ornanatan:
Sofa Beds represent a distinct functional category with unique purchase intent (dual-purpose furniture for sleeping and seating) that cannot be adequately captured by the existing style-based categories (chaise, corner, loveseat, sectional)image


ricardotejedorsanz (approved):
Let's just revert the changes to dev as these are managed separately. The rest looks great.

PR #824, opened 2025-12-08
Migrate localization files and Config to Translation Platform format

larry-perry:
Before Merging
- [x] ‼️ https://github.com/Shopify/translation-platform/pull/8832 is merged

### What

Migrate product taxonomy localization files & config from Argo to Translation Platform (TP).

### Changes

- Convert `context` data keys to YAML comments for compatibility with TP human translators
```
aa-1-1:
  # Apparel & Accessories > Clothing > Activewear
  name: Activewear
```
       
- Remove `context` keys from all target translation files

- Replace `argo_translation.yml` with `translation.yml`
   - Uses "product" translation profile

### Why

Translation Platform requires a different format for context metadata. Context should be available to human translators as comments rather than structured data that gets translated.

### Testing

- Tests pass
- Regenerated English files with new format
- Spot-checked migrated translation files

### DIFF
--- a/dev/lib/product_taxonomy/commands/sync_en_localizations_command.rb
+++ b/dev/lib/product_taxonomy/commands/sync_en_localizations_command.rb
@@ -51,6 +51,11 @@ module ProductTaxonomy
       yaml_output = YAML.dump(localizations, line_width: -1)
       yaml_with_comments = inject_context_comments(yaml_output, context_map, type)
 
+      # Verify that comment injection didn't alter the data structure
+      unless YAML.load(yaml_with_comments) == localizations
+        raise "Failed to safely inject comments into #{type} localizations"
+      end
+
       File.open(file_path, "w") do |file|
         file.puts "# This file is auto-generated. Do not edit directly."
         file.write(yaml_with_comments)


translation.yml
```diff
@@ -1,19 +1,22 @@
 source_language: en
 target_languages: [bg-BG, cs, da, de, el, es, fi, fr, hr-HR, hu, id-ID, it, ja, ko, lt-LT, nb, nl, pl, pt-BR, pt-PT, ro-RO, ru, sk-SK, sl-SI, sv, th, tr, vi, zh-CN, zh-TW]
+async_pr_mode: per_pr
+component_prefix: product-taxonomy
+translation_profile: product
```
larry-perry: cc @Aleks-Ignjatovic

translation.yml
```diff
@@ -1,19 +1,22 @@
 source_language: en
 target_languages: [bg-BG, cs, da, de, el, es, fi, fr, hr-HR, hu, id-ID, it, ja, ko, lt-LT, nb, nl, pl, pt-BR, pt-PT, ro-RO, ru, sk-SK, sl-SI, sv, th, tr, vi, zh-CN, zh-TW]
+async_pr_mode: per_pr
+component_prefix: product-taxonomy
```
ptmx, line 4: Do we need to import all of this content into Smartling before shipping this?

dev/lib/product_taxonomy/commands/sync_en_localizations_command.rb
```diff
@@ -41,13 +41,74 @@ def sync_values
...
     end
+
+    # Extracts and removes 'context' keys from the localization hash.
+    #
+    # Expects a specific structure from LocalizationsSerializer:
+    #   { "en" => { "section" => { "entry-id" => { "name" => "...", "context" => "..." } } } }
+    #
+    # Note: This creates a flat map keyed by entry ID only, so it assumes entry IDs are unique
+    # across all sections. In practice, each file (categories/attributes/values) contains only
+    # one section, so there are no collisions.
```
ptmx, line 73: Worth a `raise` if a given id already exists in the map when we process it in case any assumptions change here?

dev/lib/product_taxonomy/commands/sync_en_localizations_command.rb
```diff
@@ -41,13 +41,74 @@ def sync_values
...
+
+          if is_entry_id && context_map.key?(entry_id)
+            field_indent = next_line[/^(\s+)/, 1]
+            result << "#{field_indent}# #{context_map[entry_id]}\n"
+          end
+        end
+      end
+
+      result.join
+    end
```
danielpgross, line 128: This approach using regex matching based on indentation level strikes me as brittle and hard to follow. Are there any alternative approaches that achieve the same goal with more clarity/robustness?

larry-perry: Great question! I explored a few alternatives:

Custom YAML writer. This was my first approach, but I found that handling YAML edge cases (strings with colons, quotes, boolean-like values like "NO", etc.) made the code far more brittle and complex than the current approach.

Hardcoded indentation assumptions. I tried a similar approach to this, but without the regex by hardcoding the expected location of context comments (e.g., assume entries are always at 4 spaces, fields at 6). I thought this could be clearer, but I worried it would break if YAML.dump's formatting ever changed across Ruby versions.

I finally decided to go with the current approach and tried to address the clarity/robustness tradeoff with thorough tests and comments. I'm not married to it by any means.  If you prefer one of the other methods I explored (or have another approach in mind), I'm open to changing it. However, I'd consider the next best alternative (hardcoded indentation) to actually be more brittle, not less.

dev/lib/product_taxonomy/commands/sync_en_localizations_command.rb
larry-perry: Good call. I've added a check for duplicate entry IDs to the existing test that validates the real taxonomy data structure. I've been using tests to verify our assumption against the actual data during tests to keep this part of the code base simple.  Let me know if you prefer some kind of runtime validation step.

dev/lib/product_taxonomy/commands/sync_en_localizations_command.rb
larry-perry: I've added more comments to make the code easier to review.

dev/lib/product_taxonomy/commands/sync_en_localizations_command.rb
```diff
@@ -41,13 +41,85 @@ def sync_values
...
 
+    # Writes localization data to a YAML file with 'context' keys converted to comments.
+    # Uses YAML.dump for formatting, then injects context as comments above entry fields.
     def write_localizations(type, localizations)
       file_path = File.expand_path("localizations/#{type}/en.yml", ProductTaxonomy.data_path)
+
+      context_map = extract_contexts(localizations)
+
+      yaml_output = YAML.dump(localizations, line_width: -1)
+      yaml_with_comments = inject_context_comments(yaml_output, context_map, type)
```
sjahu, line 52: The comment injection looks legit to me but as noted below this kind of thing can be kind of brittle. Would it be worthwhile to assert that the resulting data is identical after we inject the comments?

something like this:

`raise "failed to safely inject comments" unless YAML.load(yaml_with_contents) == localizations`

translation.yml
sjahu: yes, definitely, or at least right after shipping it. I believe rob's tracking that 👍

dev/lib/product_taxonomy/commands/sync_en_localizations_command.rb
danielpgross: Thanks for sharing all that context and adding comments 👍 I can't see a better approach, so I'm on board with this one.

PR #853, opened 2026-01-29
Simplify ProductTaxonomy::Loader API to use data_path

hadijafar:
# Simplify ProductTaxonomy::Loader interface

This PR simplifies the `ProductTaxonomy::Loader.load` method by accepting a single `data_path` parameter instead of requiring individual file paths. The loader now constructs all necessary file paths internally based on the provided data directory.

The change reduces code duplication across the codebase by centralizing the file path construction logic in the loader class itself. This makes the API more straightforward to use and ensures consistent file path handling.

Version bumped to 1.1.0 to reflect this interface change.

[look at this pr for more context](https://github.com/shop/world/pull/389431)

### DIFF
--- a/dev/lib/product_taxonomy/loader.rb
+++ b/dev/lib/product_taxonomy/loader.rb
@@ -6,10 +6,9 @@ module ProductTaxonomy
   class Loader
     class << self
       def load(data_path:)
-        return if ProductTaxonomy::Category.all.any?
-
-        # Set the module-level data_path so localizations and other files can be accessed
         ProductTaxonomy.data_path = data_path
+
+        return if ProductTaxonomy::Category.all.any?
 
         values_path = File.join(data_path, "values.yml")
         attributes_path = File.join(data_path, "attributes.yml")
@@ -19,10 +18,7 @@ module ProductTaxonomy
         begin
           ProductTaxonomy::Value.load_from_source(YAML.load_file(values_path))
           ProductTaxonomy::Attribute.load_from_source(YAML.load_file(attributes_path))
-
-          if File.exist?(return_reasons_path)
-            ProductTaxonomy::ReturnReason.load_from_source(YAML.load_file(return_reasons_path))
-          end
+          ProductTaxonomy::ReturnReason.load_from_source(YAML.load_file(return_reasons_path))
 
           categories_source_data = categories_glob.each_with_object([]) do |file, array|
             array.concat(YAML.safe_load_file(file))


hadijafar:
* **#853** <a href="https://app.graphite.com/github/pr/Shopify/product-taxonomy/853?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/product-taxonomy/853?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

dev/lib/product_taxonomy/loader.rb
```diff
@@ -5,14 +5,24 @@
...
+        attributes_path = File.join(data_path, "attributes.yml")
+        return_reasons_path = File.join(data_path, "return_reasons.yml")
+        categories_glob = Dir.glob(File.join(data_path, "categories", "*.yml"))
+
         begin
           ProductTaxonomy::Value.load_from_source(YAML.load_file(values_path))
           ProductTaxonomy::Attribute.load_from_source(YAML.load_file(attributes_path))
 
-          if return_reasons_path && File.exist?(return_reasons_path)
+          if File.exist?(return_reasons_path)
```
danielpgross: I think we can actually just directly call `ProductTaxonomy::ReturnReason.load_from_source` without the surrounding `if` -- we can assume that return reasons should exist like any other resource (attributes, values)

dev/lib/product_taxonomy/loader.rb
```diff
@@ -5,14 +5,24 @@
 module ProductTaxonomy
   class Loader
     class << self
-      def load(values_path:, attributes_path:, categories_glob:, return_reasons_path: nil)
+      def load(data_path:)
+        # Set the module-level data_path so localizations and other files can be accessed.
+        # This must happen before the early return so callers can rely on data_path being set
+        # even when reloading after a reset.
+        ProductTaxonomy.data_path = data_path
```
danielpgross: The early return is a simple check to see if the taxonomy has already been loaded into memory and, if so, to avoid doing so again. It's a performance optimization.

The variable assignment is not performance sensitive so it doesn't matter whether it comes before or after the early return, but I think the comment is incorrect -- it should be fine to have it after the early return.

Let's keep as-is but just remove the comment to avoid confusion.

```suggestion
        ProductTaxonomy.data_path = data_path
```

danielpgross (approved):
Changes look good 👍 

You've got some CI failures in `GenerateDocsCommandTest` blocking the merge, you probably just need to set up some additional mocks there.

Those should be trivial to add, so I'm approving in advance.

PR #860, opened 2026-02-10
Remove command requires from product_taxonomy.rb

hadijafar:
### TL;DR

Removed command file imports from the product taxonomy library.

### What changed?

Removed all command-related import statements from the `product_taxonomy.rb` file. This includes various commands such as generate distribution, find unmapped categories, generate docs, dump data, sync localizations, and add various taxonomy elements.

### How to test?

1. Verify that the product taxonomy functionality still works correctly without these imports
2. Ensure that any code that previously used these commands is updated to import them directly
3. Run the test suite to confirm no regressions

### Why make this change?

The shopify_product_taxonomy gem 1.1.0 has a packaging issue that causes a LoadError when used as an installed gem. The gemspec intentionally excludes the commands/ directory from the published gem package (via the reject filter on spec.files), since those CLI commands are only needed for local development in the product-taxonomy repo. However, lib/product_taxonomy.rb which is the gem's main entrypoint unconditionally requires all of the command files (lines 59-75, require_relative "product_taxonomy/commands/command"). This works fine when running inside the product-taxonomy repo where those files exist on disk, but fails when the gem is installed by a consumer because the files were excluded from the package. The fix would be to either move those requires into a separate file like cli.rb that only the dev Rakefile loads, or guard them behind a file existence check so they're skipped when running as a gem.


hadijafar:
* **#860** <a href="https://app.graphite.com/github/pr/Shopify/product-taxonomy/860?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/product-taxonomy/860?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

danielpgross (approved):
Good catch, looks like these got added by accident when we merged return reasons a few weeks ago

PR #866, opened 2026-02-18
Add "pilling" return reason

akdotcom:
Merchant has asked if "pilling" reason could be added. This seems to be a valid return reason for which there is not a good matching return reason.

### DIFF
--- a/data/localizations/return_reasons/en.yml
+++ b/data/localizations/return_reasons/en.yml
@@ -821,6 +821,9 @@ en:
     pigmentation:
       name: Pigmentation
       description: The intensity or density of color in the cosmetic product is unsatisfactory.
+    pilling:
+      name: Pilling
+      description: The fabric prematurely forms unwanted fiber balls or fuzz during use or washing.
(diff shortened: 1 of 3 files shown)


ricardotejedorsanz:
Thanks @akdotcom! Is this a standalone reason or do we want to map it to a specific category like Sweaters?
cc @jennamsw @caitlinmullen

caitlinmullen:
I actually added **[pilling](https://github.com/Shopify/product-taxonomy-internal/commit/c503bae6a4f4a73843910983f2be892b64f8f44b)** a couple weeks ago but I don't think it's been merged yet.

ricardotejedorsanz:
Keeping this return reason as standalone for now (not mapped to a taxonomy category) as confirmed with @jennamsw offline!

ricardotejedorsanz:
Updated the description as suggested by @jennamsw to align with existing pattern: "The fabric prematurely forms unwanted fiber balls or fuzz during use or washing."

PR #880, opened 2026-03-02
Add "delivery issue" return reason

ricardotejedorsanz:
Add a new return reason "Delivery issue" to the taxonomy.

- **Name:** Delivery issue
- **Description:** The shipment couldn't be delivered.
- **friendly_id:** `delivery_issue`


jennamsw (approved):
Thanks for adding this!

PR #931, opened 2026-06-03
2026-05 - Add category-specific return reasons

ricardotejedorsanz:
> **Stacked on #930** (generic return reasons backfill). Base is the generics branch; retarget to `main` once #930 merges.

## What

Applies **category-specific** return reasons to 1,520 categories, sourced from the Feb 12 human + expert review batches (Electronics, Health & Beauty, and miscellaneous verticals).

Specific reasons are placed **before** the six generic reasons, matching the ordering already used across the taxonomy. Net change: **+3,907** specific assignments, with **39** stale specifics removed where the reviewed set was authoritative (e.g. a few apparel categories).

## Affected verticals

| Vertical | Categories updated |
|---|---:|
| Electronics | 577 |
| Home & Garden | 561 |
| Health & Beauty | 298 |
| Apparel & Accessories | 44 |
| Food, Beverages & Tobacco | 16 |
| Sporting Goods | 11 |
| Business & Industrial | 8 |
| Arts & Entertainment | 2 |
| Software | 2 |
| Furniture | 1 |
| **Total** | **1,520** |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/data/categories/so_software.yml
+++ b/data/categories/so_software.yml
@@ -832,6 +832,8 @@
   - video_game_platform
   - video_game_sub_genre
   return_reasons:
+  - region
+  - incompatible
   - changed_my_mind
   - item_not_as_described
   - received_the_wrong_item
@@ -851,6 +853,7 @@
   - video_game_platform
   - video_game_sub_genre
   return_reasons:
+  - incompatible
   - changed_my_mind
   - item_not_as_described
   - received_the_wrong_item
--- a/data/categories/aa_apparel_accessories.yml
+++ b/data/categories/aa_apparel_accessories.yml
@@ -8554,6 +8554,7 @@
   - too_small
   - color
   - style
+  - material
   - changed_my_mind
   - item_not_as_described
   - received_the_wrong_item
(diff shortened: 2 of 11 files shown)


data/categories/so_software.yml
```diff
@@ -832,6 +832,8 @@
   - video_game_platform
   - video_game_sub_genre
   return_reasons:
+  - region
```
Waterfountain10, line 835: what does region mean? incompatible region for the user? or not enough regions in general. Region-locking reads as a form of incompatibility, so these two will likely get picked interchangeably, no?

if its different, should it not be a little bit more specific?

Waterfountain10 (changes requested):
i added a comment ^ worth a look

data/categories/so_software.yml
ricardotejedorsanz: Good catch! Will flag to the Orders & Returns team for an update downstream, but in the meantime the description should make intent clearer: `The service or product is locked to a geographic area different from the customer's location.`

vincehouhou (approved):

Waterfountain10 (approved):

PR #940, opened 2026-06-08
Release v2026-05

KaiKitJeffreyChan:
### Tl;dr
The final v2026-05 release.

### How it was generated
I generated the release using `bin/product_taxonomy release` 

The following files now return git LFS pointers due to Githubs 100MB file threshold 
- dist/th/taxonomy.json 
- dist/el/taxonomy.json  
- dist/bg-BG/taxonomy.json
- dist/ru/taxonomy.json
- dist/th/categories.json

Used `git lfs migrate import --above=100MiB` command to rewrite any file in commit history that has files that exceeded this threshold. There were only files of this size introduced in these two commits but this command can be used in the future when files exceed 100MiB. Edits to `.gitattributes` is needed to ensure future edits to git LFS files are not overwritten, and the contents are correctly mapped to a pointer (otherwise it will rewrite the file)

### How to review
The diff is too big to view on Github. Instead, you can checkout the branch locally and look at each commit individually. For each commit, you can:

- browse around the repository to confirm files look as expected
- use git diff or git difftool to see changes


danielpgross (changes requested):
I compared with the `v2026-02` release and found 2 differences:

1. **Stale google mapping.** The "bump to next-unstable" commit adds the new `shopify_2026-08_to_google` mapping but keeps the old `shopify_2026-05_to_google` one, leaving two google mapping files. In the last two cycles the bump replaced it cleanly (always one file). Let's remove the old one.
2. **Localized `attributes.txt` missing their header.** In this release, every `dist/<locale>/attributes.txt` lost its `"# Shopify Product Taxonomy - Attributes: <version>"` header. `en/attributes.txt` still has it, and the sibling files (`categories.txt, attribute_values.txt, return_reasons.txt`) still have it for all locales. So only localized `attributes.txt` is affected, and it's inconsistent with everything else. Looks like an unintended regression in the localized attributes export.

danielpgross (approved):
Explored the diff locally, checked all `data` changes, spot-checked `dist` for a few locales. LGTM 👍

PR #942, opened 2026-06-16
Fix AlphanumericSorter crash on mixed sequential-text values

KaiKitJeffreyChan:
## What

`bin/product_taxonomy dist/release` crashes with `ArgumentError: comparison of Array with Array failed` (in `Value.sort_by_localized_name`) when generating the `da` locale files for the new `ev_connector_adapter_direction` attribute. Latent sorter bug, not a data problem — translations are complete.

## Root cause

`AlphanumericSorter#normalize_sequential` builds three sort-key slots straight from regex captures with no fallback (`primary_text`, `secondary_text`, `trailing_text`). The range capture only matches when a number follows the separator, so within one attribute, names with digits parse differently from names without:

| `da` name | `secondary_text` slot |
| --- | --- |
| `CCS1-bilindtag til CHAdeMO-stik` | `nil` |
| `CCS1-bilindtag til J1772-stik` | `"bilindtag til j"` |

`sort_by` then compares a `nil` slot against a `String`: `nil <=> "..."` → `nil`, so `Array#<=>` returns `nil` → `ArgumentError`. `en`/`fr` land `nil` in that slot uniformly, so they never crash.

## Fix

Add the `|| ""` fallback the numeric branch (`normalize_numerical`) already uses to the three text slots. Only `secondary_text` can be `nil` with the current pattern — the other two fallbacks are defensive. Every slot is now always a `String`/number, so `<=>` can't fail. Sort order is unchanged for working cases (a `nil` slot was never sortable); `dist/` output is identical.

## Tests

`AlphanumericSorter` had none. Adds a unit test file — including an invariant test that every sort-key slot is a `String` or `Numeric` across 12 name shapes — plus an end-to-end `Value.sort_by_localized_name` test using the real `da` EV-connector records from `data/values.yml`.

- Full `dev` unit suite green (432 runs, 0 failures).
- Reverting only the source fix: the sorting tests reproduce the exact `ArgumentError`, and the invariant test pinpoints the `nil` slot.

### DIFF
--- a/dev/lib/product_taxonomy/alphanumeric_sorter.rb
+++ b/dev/lib/product_taxonomy/alphanumeric_sorter.rb
@@ -37,13 +37,13 @@ module ProductTaxonomy
 
       def normalize_sequential(match)
         [
-          normalize_text(match[:primary_text]),
+          normalize_text(match[:primary_text]) || "",
           normalize_single_number(match[:primary_step]),
           normalize_text(match[:primary_unit] || match[:secondary_unit]) || "",
           normalize_text(match[:seperator]) || "-",
-          normalize_text(match[:secondary_text]),
+          normalize_text(match[:secondary_text]) || "",
           normalize_single_number(match[:secondary_step]),
-          normalize_text(match[:trailing_text]),
+          normalize_text(match[:trailing_text]) || "",
         ]
       end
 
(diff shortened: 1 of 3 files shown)


jmignac (approved):

hadijafar (approved):
looks good to me

PR #952, opened 2026-07-01
[UoM prototype] Render measurement attributes in docs (3/4)

ricardotejedorsanz:
## What this PR does

This PR updates the public docs/visualizer path to understand measurement attributes introduced in #950 and bulk-imported in #951.

It is stacked on #951 so the visualizer can be reviewed with the V0 measurement vocabulary.

## What changed

- Docs attribute serializers now emit `type` for attributes.
- Measurement attributes emit `measurement_type` and `supported_units` instead of a values list.
- Category and attribute visualizer pages render each attribute's type and measurement metadata instead of assuming every attribute has closed-list values.
- The root README documents the V0 attribute-type shape.

## UX changes

The visualizer now makes the attribute type explicit when an attribute is expanded:

- Closed-list attributes show `Type: closed_list`, followed by a `Values` label and their allowed taxonomy values.
- Measurement attributes show `Type: measurement`, `Measurement type`, and `Supported units` instead of a values list.
- The dedicated attribute search/details page uses the same measurement metadata treatment, so measurement attributes can be understood outside of a category context as well.

Image 1 — closed-list attribute expanded in a category:

image

Image 2 — measurement attribute expanded in a category:

image

Image 3 — measurement attribute on the attribute details page:

image

## Validation

- `ruby -c` passed for the changed Ruby serializers/tests.
- Focused docs serializer tests passed.
- `bundle exec rake test:unit` passed.
- `bundle exec bin/product_taxonomy docs` passed.
- `bundle exec jekyll build --source ../docs --destination /tmp/product-taxonomy-docs-site` passed.
- Verified generated unstable docs data and rendered HTML include measurement attributes with `measurement_type` and `supported_units`, and closed-list attributes with a `Values` label.
- `git diff --check` passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@
   - [🕹️ Interactive explorer](#️-interactive-explorer)
 - [🧭 Getting started](#-getting-started)
   - [🧩 1. Integrators: How to integrate with the taxonomy](#-1-integrators-how-to-integrate-with-the-taxonomy)
+    - [Attribute types](#attribute-types)
     - [🗺️ Mapping to other taxonomies](#️-mapping-to-other-taxonomies)
   - [🧑🏼‍🏫 2. Taxonomists: How to make changes to the taxonomy](#-2-taxonomists-how-to-make-changes-to-the-taxonomy)
   - [👩🏼‍💻 3. Developers: How to evolve the system](#-3-developers-how-to-evolve-the-system)
@@ -51,6 +52,10 @@ You can think of this repository serving 3 primary users:
 Dive straight into [`dist`](./dist/) to find the files you need and integrate this taxonomy into your system.
 
 We offer `txt` and `json` formats to make it easy to integrate with your systems. If you have a specific format you'd like to see, please open an issue and let us know!
+
+#### Attribute types
+
+Attributes include an explicit `type`. Closed-list attributes define a set of taxonomy values. Measurement attributes define a `measurement_type` and `supported_units` instead of values.
 
 #### 🗺️ Mapping to other taxonomies
 
--- a/README.md
+++ b/README.md
@@ -55,7 +55,41 @@ We offer `txt` and `json` formats to make it easy to integrate with your systems
 
 #### Attribute types
 
-Attributes include an explicit `type`. Closed-list attributes define a set of taxonomy values. Measurement attributes define a `measurement_type` and `supported_units` instead of values.
+Attributes include an explicit `type`:
+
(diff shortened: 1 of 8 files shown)


jeanguo-png (commented):
Non-blocking: could we default missing `attribute.type` to `closed_list` in the shared layouts? Older release docs data does not include `type`, so those pages will render a blank Type value even though the attributes are effectively closed-list. A small Liquid default would keep old release pages looking correct while supporting measurement attributes in the new data.

README.md
```diff
@@ -52,6 +53,10 @@ Dive straight into [`dist`](./dist/) to find the files you need and integrate th
 
 We offer `txt` and `json` formats to make it easy to integrate with your systems. If you have a specific format you'd like to see, please open an issue and let us know!
 
+#### Attribute types
+
+Attributes include an explicit `type`. Closed-list attributes define a set of taxonomy values. Measurement attributes define a `measurement_type` and `supported_units` instead of values.
```
danielpgross: This is pretty terse, it might help to add some examples here for clarity

danielpgross (approved):
Approach looks fine (details will depend on whether we make further changes to the data model)

ricardotejedorsanz:
> Non-blocking: could we default missing attribute.type to closed_list in the shared layouts? Older release docs data does not include type, so those pages will render a blank Type value even though the attributes are effectively closed-list.

Thanks @jeanguo-png, addressed this in a separate follow-up commit.

README.md
ricardotejedorsanz: Thanks @danielpgross, I included one-line definitions for `closed_list` and `measurement`, plus compact JSON examples for both attribute shapes:

#### Attribute types

Attributes include an explicit `type`:

- `closed_list`: an attribute whose value is selected from a predefined set of taxonomy values, used when consistent value standardization is needed across products.
- `measurement`: an attribute whose value is a number paired with a supported unit, used when products need precise measurements such as dimensions, weight, volume, or power.

Closed-list attributes include `values`:

```json
{
  "id": "gid://shopify/TaxonomyAttribute/1",
  "name": "Color",
  "handle": "color",
  "type": "closed_list",
  "values": [
    {
      "id": "gid://shopify/TaxonomyValue/1",
      "name": "Black",
      "handle": "color__black"
    }
  ]
}
```

Measurement attributes include `measurement_type` and `supported_units` instead of `values`:

```json
{
  "id": "gid://shopify/TaxonomyAttribute/12429",
  "name": "Height",
  "handle": "height",
  "type": "measurement",
  "measurement_type": "dimension",
  "supported_units": ["cm", "in"]
}
```

PR #954, opened 2026-07-02
Fix Shopify mapping release finalization

JeffrayZhang:
## Summary

This PR fixes release mapping finalization so stable releases only rewrite Shopify `to_shopify.yml` mappings that are still pending finalization.

Related issue: shop/issues-taxonomy#474

## Context

The release command previously selected the `to_shopify.yml` file to finalize with `files.sort.last`. That assumes the newest Shopify version directory always has its own `to_shopify.yml`, but releases only produce one when that cycle has remappings.

When an intermediate release has no mapping file, the path-order heuristic can select an older, already-finalized mapping and rewrite its `output_taxonomy` to the new release. That skips the intermediate version in the mapping chain and breaks the mapping validation test.

## Changes

- Replaced `files.sort.last` selection with state-based filtering: only `to_shopify.yml` files whose `output_taxonomy` still points at `shopify/YYYY-MM-unstable` are finalized.
- Updated the existing `to_shopify.yml` special-case test so finalized mappings stay untouched.
- Added a regression test through `execute` for the no-pending-mapping case, preserving already-finalized Shopify links.

## Testing

- `bundle exec ruby -Itest test/commands/generate_release_command_test.rb`
- `bundle exec rubocop lib/product_taxonomy/commands/generate_release_command.rb test/commands/generate_release_command_test.rb`

I also ran `bundle exec rake test` locally with an isolated Ruby 4 bundle workaround. Unit and integration tests passed; the existing benchmark task failed on a performance threshold in this local Ruby 4/Nix environment. Full CI will be triggered with `devx ci --full`.

### DIFF
--- a/dev/lib/product_taxonomy/commands/generate_release_command.rb
+++ b/dev/lib/product_taxonomy/commands/generate_release_command.rb
@@ -132,12 +132,17 @@ module ProductTaxonomy
 
     def update_mapping_files(filename, field_name, new_value)
       files = Dir.glob(File.expand_path("integrations/**/mappings/#{filename}", ProductTaxonomy.data_path))
+      escaped_field_name = Regexp.escape(field_name)
 
-      files = [files.sort.last] if filename == "to_shopify.yml"
+      if filename == "to_shopify.yml"
+        files = files.select do |file|
+          File.read(file).match?(%r{#{escaped_field_name}: shopify/\d{4}-\d{2}-unstable})
+        end
+      end
 
       files.each do |file|
         content = File.read(file)
-        content.gsub!(%r{#{field_name}: shopify/\d{4}-\d{2}(-unstable)?}, "#{field_name}: #{new_value}")
+        content.gsub!(%r{#{escaped_field_name}: shopify/\d{4}-\d{2}(-unstable)?}, "#{field_name}: #{new_value}")
         File.write(file, content)
       end
     end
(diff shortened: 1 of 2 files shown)


dev/lib/product_taxonomy/commands/generate_release_command.rb
```diff
@@ -133,7 +133,11 @@ def update_integration_mappings_next_unstable
     def update_mapping_files(filename, field_name, new_value)
       files = Dir.glob(File.expand_path("integrations/**/mappings/#{filename}", ProductTaxonomy.data_path))
 
-      files = [files.sort.last] if filename == "to_shopify.yml"
+      if filename == "to_shopify.yml"
+        files = files.select do |file|
+          File.read(file).match?(%r{#{field_name}: shopify/\d{4}-\d{2}-unstable})
```
KaiKitJeffreyChan: nit: Can we use `Regep.escape` to make `field_name` explict

KaiKitJeffreyChan (approved):
Nice! thanks for taking this task 😄

dev/test/commands/generate_release_command_test.rb
```diff
@@ -158,6 +158,21 @@ class GenerateReleaseCommandTest < TestCase
       assert_equal "input_taxonomy: shopify/#{@next_version}", from_shopify_content
     end
 
+    test "execute preserves finalized to_shopify mapping when release has no pending mapping" do
```
jmignac, line 161: The tests cover multiple `to_shopify.yml` files being iterated, and the non-pending ones being skipped.  Is it possible that two or more files are in the `pending` state at the same time? 🤔 

If so, it would be nice to ahve a test for that too, to lock that in against future regressions. Otherwise it looks good to me

jmignac (approved):

PR #956, opened 2026-07-06
Add disclosure taxonomy axis

ruba-ruba:
### TL;DR
Adds a new **`disclosures`** axis to the taxonomy — legally-required product disclosures (safety warnings, chemical-exposure notices) modeled as a shallow hierarchy of grouping + leaf nodes, following the same pattern as the existing `return_reasons` axis.

### Why
Shopify's product-disclosure feature needs its taxonomy authored and versioned in the open, alongside categories and return reasons, so contributions follow the same public PR flow and downstream systems consume it from one source of truth instead of an internal fork.

### What
- **`data/disclosures.yml`** — the axis source data (10 nodes: `choking_hazard` and `chemical_exposure` groups + 8 leaf disclosures spanning US CPSC choking hazards and CA Prop 65 warnings).
- **`ProductTaxonomy::Disclosure`** — model with `load_from_source`, `Indexed` lookups (`.all`/`find_by`), `gid`, hierarchy helpers (`parent`/`children`/`leaf?`/`root?`), and validations:
  - stable integer `id`, unique `public_id`, required `name`/`internal_label`;
  - `jurisdictions` and `display_preferences` allowed **only** on leaf nodes;
  - `display_preferences.surfaces` must be a unique subset of `product_page`, `cart`, `checkout`;
  - `parent_id` referential integrity.
- **Loader wiring** — loads `disclosures.yml` and runs the hierarchy-dependent validations in the post-load `:taxonomy_loaded` pass.
- **Tests** — unit coverage for load/uniqueness/required-fields and every validation branch, plus an integration assertion that the real `disclosures.yml` loads consistently.

### Notes for reviewers
- Modeled on `return_reasons` (flat axis) but extended with `parent_id` for the disclosure hierarchy.
- `Disclosure` intentionally does **not** include `FormattedValidationErrors`: that mixin calls `friendly_id` on non-`Category` models, which disclosures don't have. Happy to make that mixin generic in a follow-up if preferred.
- Scoped to data + model + loader + validation. Dist/docs serializers, per-locale `dist/` artifacts, `localizations/disclosures/`, and a CUE `schema/data` file are **not** included (consistent with `return_reasons`, which ships without a CUE schema) — can follow up if the axis should reach full parity.

### Verification
- Repo unit + integration suites cover the new model and the real data file.
- Locally exercised the model + loader against the real `data/disclosures.yml` and all validation edge cases (leaf/grouping constraints, invalid/duplicate surfaces, dangling `parent_id`, duplicate `public_id`, non-integer `id`) — all green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


data/disclosures.yml
```diff
@@ -0,0 +1,219 @@
...
+- id: 2
+  public_id: us-cpsc-choking_small_parts
+  parent_id: 1
+  name: "Choking Hazard — Small Parts (US)"
+  internal_label: "Choking Hazard — Small Parts (US)"
+  description:
+  jurisdictions:
+  - US
+  legal_citation: "16 CFR 1501"
+  symbol: "https://cdn.shopify.com/static/product-disclosures/default-warning-white.png"
```
danielpgross, line 44: Since this repo is public and open for community contributions, we should think about how contributors should contribute these images. As things stand, there's no pathway for contributors to upload images to this CDN path.

data/disclosures.yml
```diff
@@ -0,0 +1,219 @@
...
+# is a grouping/root node; a node that no other node points at is a leaf.
+#
+# Authoring rules (enforced by ProductTaxonomy::Disclosure validations):
+#   - Only leaf nodes define `jurisdictions` and `display_preferences`.
+#   - Grouping/root nodes must NOT define `jurisdictions` or `display_preferences`.
+#   - `display_preferences.surfaces` values must be a unique subset of
+#     product_page, cart, checkout.
+
+# Choking hazards
+- id: 1
```
danielpgross: What advantage do we get from having a numeric ID here? If `public_id` is anyway guaranteed unique, I'd prefer to only use that for simplicity. This would line up with how categories are defined in the main product taxonomy axis.

dev/lib/product_taxonomy/models/disclosure.rb
```diff
@@ -0,0 +1,187 @@
+# frozen_string_literal: true
+
+module ProductTaxonomy
+  # A legally-required product disclosure (e.g. a safety warning or chemical
+  # exposure notice). Disclosures form a shallow hierarchy: grouping/root nodes
+  # (no `parent_id`) organize the axis, and leaf nodes carry the jurisdiction
+  # and display information that merchants surface.
+  class Disclosure
+    include ActiveModel::Validations
+    extend Indexed
```
danielpgross, line 12: I noticed we don't have `extend Localized` here, do we intend for it them be localized at some point?

dev/lib/product_taxonomy/models/disclosure.rb
```diff
@@ -0,0 +1,187 @@
+# frozen_string_literal: true
+
+module ProductTaxonomy
+  # A legally-required product disclosure (e.g. a safety warning or chemical
+  # exposure notice). Disclosures form a shallow hierarchy: grouping/root nodes
+  # (no `parent_id`) organize the axis, and leaf nodes carry the jurisdiction
+  # and display information that merchants surface.
+  class Disclosure
+    include ActiveModel::Validations
```
danielpgross, line 9: Did you consider using `FormattedValidationErrors` for nicely formatted validation errors?

dev/lib/product_taxonomy/models/disclosure.rb
```diff
@@ -0,0 +1,187 @@
...
+    validates :id, presence: true, numericality: { only_integer: true }, on: :create
+    validates :public_id, presence: true, on: :create
+    validates :name, presence: true, on: :create
+    validates :internal_label, presence: true, on: :create
+    validates_with ProductTaxonomy::Indexed::UniquenessValidator, attributes: [:public_id, :id], on: :create
+
+    # Validations that can only run once the whole axis is loaded.
+    validate :parent_reference_exists, on: :taxonomy_loaded
+    validate :jurisdiction_and_display_only_on_leaves, on: :taxonomy_loaded
+    validate :display_preferences_surfaces_are_valid, on: :taxonomy_loaded
```
danielpgross: Validations are important here because they act as the main automated quality gate for community contributions. We should consider adding more for better coverage:

1. Validate `public_id` format as a stable slug.
2. Validate hierarchy integrity: no self-parenting, no cycles, no orphaned branches.
3. Enforce intended depth: root/grouping node → leaf disclosure only.
4. Validate field types: arrays are arrays, hashes are hashes, IDs are integers, strings are strings.
5. Validate leaf completeness: required legal/display fields like jurisdiction, title/content, citation/source are present.

danielpgross (commented):
Looks good so far. Left some comments, biggest question IMO is around IDs: using public IDs only seems like it would be simpler than numeric ID + public ID.

data/disclosures.yml
ruba-ruba: removed

dev/lib/product_taxonomy/models/disclosure.rb
ruba-ruba: added

dev/lib/product_taxonomy/models/disclosure.rb
ruba-ruba: added

dev/lib/product_taxonomy/models/disclosure.rb
ruba-ruba: added more validations

data/disclosures.yml
ruba-ruba: Good question and I don't have an answer yet.
Let's wait until we get confirmation whether we'll be open to community contributions for disclosures first.

ruba-ruba: From the slack: we will think how CRN url can be provided if/when we receive confirmation that community can contribute to disclosures

dev/lib/product_taxonomy/models/disclosure.rb
```diff
@@ -0,0 +1,268 @@
...
+    # @return [Disclosure, nil]
+    def parent
+      parent_public_id && Disclosure.find_by(public_id: parent_public_id)
+    end
+
+    # The direct children of this disclosure.
+    #
+    # @return [Array<Disclosure>]
+    def children
+      Disclosure.all.select { |disclosure| disclosure.parent_public_id == public_id }
```
danielpgross, line 170: Searching all disclosures is O(n), so this will become O(n^2) when called on all disclosures (ex. during validation after loading). Fine for now while we have a small number of disclosures, but we'll probably want to optimize that before the disclosure count gets much bigger. We can solve it similar to how we do for `Category`: do a second pass during loading to cache parent/child relationships. Not a blocker.

danielpgross (approved):
Nice 👍

PR #957, opened 2026-07-13
Publish stable release distribution assets

JeffrayZhang:
### TL;DR

Adds automated tooling to stage and publish distribution assets to GitHub Releases as gzip-compressed files. View this tech design Doc: https://docs.google.com/document/d/11G8FCV5D24TdRIV9V4WUSbKskp-MUpVfXAiyoJc6hDk/edit?tab=t.0#heading=h.1evip2vrmlc4

### What changed?

- Introduces `DistAssetStager`, which reads locale-organized distribution files, flattens their paths into dot-separated asset names (e.g. `integrations/all_mappings.json` under `en/` becomes `integrations.all_mappings.en.json.gz`), gzip-compresses each file with deterministic metadata (mtime=0), and writes them to a clean output directory.
- Adds a `stage_dist_assets` CLI command that wraps `DistAssetStager` for standalone use.
- Adds a `publish_release_assets` CLI command that, given a stable release tag (e.g. `v2024-01`), checks out the tag into a temporary Git worktree, regenerates the distribution, stages the assets, uploads them to the corresponding GitHub Release via `gh`, and verifies that all expected assets appear on the release. If an upload partially succeeds, the error message includes step-by-step retry instructions.
- Updates the `generate_release` post-run summary to include the new steps: running `publish_release_assets` and spot-checking a published asset with `curl`.

### How to test?

1. Generate a release branch and follow the updated next-steps output to confirm the new steps appear correctly.
2. Run the stager directly against a local `dist/` directory:

    Inspect `/tmp/staged` to confirm gzip files are named correctly and decompress cleanly.
3. After pushing a stable tag and creating a GitHub Release, run:

    Then spot-check a published asset:
4. Run the new unit tests: `DistAssetStagerTest` and `PublishReleaseAssetsCommandTest`.

### Tophat

Ran the flow against a disposable prerelease in a fork of the public repo.

```sh
git tag v2099-01 792a47c34ed73318ca440f7b2e38972cd944598f
git push https://github.com/JeffrayZhang/product-taxonomy.git refs/tags/v2099-01
gh release create v2099-01 --repo JeffrayZhang/product-taxonomy --prerelease --title "Temporary gzip release asset tophat v2099-01"

cd dev
GH_REPO=JeffrayZhang/product-taxonomy \
  bundle exec bin/product_taxonomy publish_release_assets v2099-01

curl --fail --location \
  https://github.com/JeffrayZhang/product-taxonomy/releases/download/v2099-01/categories.en.json.gz \
  | gunzip | ruby -rjson -e 'JSON.parse(STDIN.read); puts "valid JSON"'

gh release delete v2099-01 --repo JeffrayZhang/product-taxonomy --yes --cleanup-tag
git tag -d v2099-01
```

### Why make this change?

Previously, release assets had to be manually prepared and uploaded to GitHub Releases. This automates the entire asset publication pipeline generation, compression, upload, and verification—reducing the risk of human error and making the release process more reproducible.

Closes https://github.com/shop/issues-taxonomy/issues/541.


dev/lib/product_taxonomy/dist_asset_stager.rb
```diff
@@ -0,0 +1,112 @@
...
+        destination_path
+      end
+    end
+
+    private
+
+    def validate_paths!
+      raise ArgumentError, "Input path does not exist: #{@input_path}" unless File.directory?(@input_path)
+
+      output_inside_input = @output_path == @input_path || @output_path.start_with?("#{@input_path}#{File::SEPARATOR}")
```
KaiKitJeffreyChan: So here we are checking for the case where `input = dist/` and `output = dist/release-assets/` - then we are ensuring that the output does not live within the input. However we dont do the opposite check for instance `output = repo/` and `input = repo/dist` output is deleted before compression which would also delete the input.

dev/lib/product_taxonomy/cli.rb
```diff
@@ -62,6 +64,18 @@ def release(current_version, next_version)
       GenerateReleaseCommand.new(options.merge(current_version:, next_version:)).run
     end
 
+    desc "stage_dist_assets", "Flatten and gzip distribution files using release asset names"
+    option :input_path, type: :string, required: true, desc: "Distribution directory to stage"
+    option :output_path, type: :string, required: true, desc: "Clean directory in which to stage assets"
```
KaiKitJeffreyChan: nit: 
```suggestion
    option :output_path, type: :string, required: true, desc: "Directory to replace with staged assets; existing contents will be deleted"
```

dev/lib/product_taxonomy/commands/publish_release_assets_command.rb
```diff
@@ -0,0 +1,184 @@
...
+    def verify_assets!(repository_root, staged_files)
+      uploaded_asset_output = run_command!(
+        "gh",
+        "release",
+        "view",
+        @tag,
+        "--json",
+        "assets",
+        "--jq",
+        ".assets[].name",
```
KaiKitJeffreyChan:
```suggestion
        ".assets[] | select(.state == \"uploaded\") | .name",
```

dev/lib/product_taxonomy/commands/publish_release_assets_command.rb
```diff
@@ -0,0 +1,184 @@
...
+            run_command!(
+              "git",
+              "worktree",
+              "remove",
+              "--force",
+              source_path,
+              chdir: repository_root,
+              failure_message: "Could not remove temporary worktree for #{@tag}.",
+            )
+          end
```
KaiKitJeffreyChan, line 82: If the `if` block raises an exception and the `else` block also raises an exception, the `if` block's exception will be discarded. Maybe we can do something like this so we can see the primary error.
```suggestion
          begin
            run_command!(
              "git",
              "worktree",
              "remove",
              "--force",
              source_path,
              chdir: repository_root,
              failure_message: "Could not remove temporary worktree for #{@tag}.",
            )
          rescue => cleanup_error
            logger.warn(
              "Could not remove temporary worktree at #{source_path}: #{cleanup_error.message}. " \
                "Remove it manually with `git worktree remove --force #{source_path}`.",
            )
          end
```

KaiKitJeffreyChan (commented):
Nice work on this! Every part makes sense, just a few nits and should be good to go

dev/lib/product_taxonomy/dist_asset_stager.rb
JeffrayZhang: thanks, good catch!

KaiKitJeffreyChan (approved):
Nice! This is also low risk as there CLI commands

hadijafar (approved):
Looks good to me and talked about it in tuple

PR #959, opened 2026-07-15
Remove translation config components migrated to internal repo

sjahu:
Related to https://github.com/Shopify/product-taxonomy-internal/pull/403.

### DIFF
--- a/translation.yml
+++ b/translation.yml
@@ -5,24 +5,6 @@ async_pr_mode: per_component
 component_prefix: product-taxonomy
 translation_profile: taxonomy
 components:
-  - name: attributes
-    audience: [buyer, merchant, partner]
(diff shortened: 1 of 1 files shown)


sjahu:
> ⚠️ New languages detected. After this PR is merged, backfill translation jobs will be queued automatically for:

This is not correct-- I'll look into it before merging this...

Edit: New languages were detected because the common ancestor config was invalid and treated as nil. I updated the warning message logic to not post a message when either config version is missing. Separately, the reason the common ancestor config is invalid is that I had already removed these three components from the internal list allowing use of deprecated regionalized locale codes. 🤦 I bypassed the check to allow this to merge.

ricardotejedorsanz (approved):
Perfect, thank you @sjahu !

PR #960, opened 2026-07-16
Add workflow to publish unstable distribution assets

KaiKitJeffreyChan:
### Summary

Adds a workflow that refreshes the rolling `unstable` prerelease when taxonomy data or dev changes on `main`.

- Generates distribution files for all locales.
- Reuses the staging command from #957.
- Validates asset counts and gzip integrity before modifying the release.
- Creates the prerelease when absent and replaces assets with `--clobber`.
- Reconciles published assets against the staged set after upload: deletes orphaned assets and fails if any staged asset is missing from the release.
- Serializes runs to prevent overlapping publications.
- Supports `workflow_dispatch` for manual recovery re-runs, guarded to refuse non-main refs.
- Posts a Slack message via incoming webhook when the publish job fails, so failures don't rely on the push actor's email (~83% of `data/**` commits are the translation bot).

Stacked on #967, which adds return reason localizations to the validator this workflow relies on for all-or-nothing generation.

Closes https://github.com/shop/issues-taxonomy/issues/542.

### Known gaps (acknowledged)

- **`--clobber`** **alone can't remove or verify assets** — it only adds/replaces name-matched assets. The reconcile step closes this. Asset replacement is still non-atomic (GitHub has no transactional multi-asset swap), so a brief mixed window exists mid-run; every green run converges to the exact staged set.
- **Translation lag turns runs red** — English `data/**` changes land days before bot translations; the localization validator fails the publish during that window. Accepted: fail-loud is intended, assets stay at last-good, and the run self-heals once translations merge. Will also request that data changes arrive alongside localizations.
- **Missing localized descriptions silently fall back to English** — the validator gates `name` only; a translated entry lacking a `description` ships English text with a green run. Accepted as fallback-by-design; verified zero occurrences across all ~280k localized attribute/return-reason entries today.
- **CI generates dist in English only** — locale-specific generation bugs (the #942 class) surface first in this workflow, not CI. Accepted: this workflow is the all-locale canary; failures alert via Slack and recover via `workflow_dispatch` re-run.


KaiKitJeffreyChan:
* **#960** <a href="https://app.graphite.com/github/pr/Shopify/product-taxonomy/960?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/product-taxonomy/960?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#967** <a href="https://app.graphite.com/github/pr/Shopify/product-taxonomy/967?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

jmignac (approved):

hadijafar (approved):
looks good

PR #974, opened 2026-08-10
Port size_type localizations from Translation Platform output

KaiKitJeffreyChan:
### Summary

Ports the Translation-Platform-produced localizations for the `size_type` attribute and its 7 values into all 30 locale files. This closes the only localization gap in this repo's `data/` (verified by a key-level audit of every EN key × every locale across attributes, categories, values, and return_reasons).

- `data/localizations/attributes/<locale>.yml`: adds `size_type` (`name` + `description`) × 30 locales
- `data/localizations/values/<locale>.yml`: adds `size_type__big/maternity/other/petite/plus/regular/tall` (`name`) × 30 locales
- 60 files, +510 lines, no deletions, data-only

### Why

`size_type` landed in `92313e71` (Jul 22) with English only. Since the July translation migration, this repo's `translation.yml` covers only `return_reasons`, so translation-platform[bot] delivered these strings to the internal taxonomy repo instead — and the internal→public sync is not yet operational, leaving them stranded.

Without this, `generate_dist --locales all` fails: `GenerateDistCommand` runs `LocalizationsValidator.validate!`, which raises `Missing or incomplete localizations for the following keys ... size_type` for every non-EN locale. That blocks the unstable-assets workflow in #960.

### Provenance — no strings authored in this PR

Every string is copied **verbatim** from translation-platform[bot] output in `Shopify/product-taxonomy-internal`:

| Content | Source ref | Commit |
|---|---|---|
| attributes (30 locales) | `translation-platform/async/main/attributes` | `3d129eb` |
| values (28 locales) | `translation-platform/async/main/values` | `c63b3cd` |
| values `es`/`ja` | PR #409 head (bot-authored, delivered before branch regeneration) | `9ab6a17` |

### Verification

- Byte-equality: all 270 ported strings compared equal to their TP source values after insertion
- Completeness: every key in `attributes/en.yml` and `values/en.yml` now present in all 30 locale files (the exact check `LocalizationsValidator` performs)
- All 60 modified files re-parsed as valid YAML

Precedent for maintainer-authored localization-file PRs: #929 / the Jun 8 ES fix.

---
_Generated with Pi._


data/localizations/values/zh-CN.yml
```diff
@@ -43943,6 +43943,20 @@ zh-CN:
...
+    size_type__other:
+      name: 其他
+    size_type__petite:
+      name: 娇小
+    size_type__plus:
+      name: 加大
+    size_type__regular:
+      name: 常规
+    size_type__tall:
+      name: 加长
```
hadijafar, line 43959: Big/Plus values appear reversed? Not entirely sure but pi flagged it as something that doesnt translate correctly. However this would mainly on the data folks so shouldnt be a problem

hadijafar (approved):
Looks good, nice data port

JeffrayZhang (approved):

data/localizations/values/zh-CN.yml
KaiKitJeffreyChan: I think this should be fine, the translations were ported directly over from the translation platform. However worth a shout, I will message in #help-localizations to ask about this

PR #982, opened 2026-08-14
Maintain `return_reason` sorting when serializing categories

rachel-carvalho:
The sorting of the return reasons is meaningful so we need to maintain them, apart from `unknown` and `other_reason` which needs to stay at the end.

This removes the call to alphabetize them from the Category serialization class.

It also fixes the special treatment of `other_reason` which is using the incorrect `other` friendly id.

PR #983, opened 2026-08-14
Reflect inherited return reasons in the category YML

rachel-carvalho:
Categories can now declare that they inherit their return reasons from an ancestor instead of listing them out. This keeps return reason inheritance an explicit, data-driven concept: the source data only spells out return reasons where they actually change, and every generated artifact still contains the fully materialized set.

### Changes

**New `return_reasons: inherit` form on `Category`**
- In the category data, `return_reasons` is now either an explicit list of return reason friendly IDs or the literal value `inherit`
- A category marked `inherit` copies the return reasons of the closest ancestor that defines its own list (i.e. the nearest ancestor whose `return_reasons` is an explicit array)
- Inheritance is resolved on load, after the category tree is assembled, and each inheriting category gets its own copy of the reasons
- `Category` exposes an `inherits_return_reasons` reader so callers can tell whether a category inherits or defines its own reasons
- Root categories cannot be marked `inherit` since there is no ancestor to inherit from

**Global return reasons**
- `ReturnReason` defines a baseline set of global reasons (`changed_my_mind`, `item_not_as_described`, `received_the_wrong_item`, `damaged_or_defective`, `unknown`, `other_reason`) via `GLOBAL_FRIENDLY_IDS` and the `ReturnReason.global` accessor
- Categories that define their own list but end up with none get the global reasons as a fallback (`apply_default_return_reasons`), so every category always has at least the baseline set
- Global reasons are always ordered at the end of a category's list

**Serialization**
- The data serializer round-trips the source format: inheriting categories are written back as `return_reasons: inherit`, and defining categories keep their explicit (ordered) list
- The distribution and docs serializers read the resolved reasons, so inheriting categories emit the full, materialized return-reason set, nothing needs to be pre-expanded in the source data

**`add_category` defaults to inheriting**
- New categories default to `inherit`, so a newly added category picks up its ancestor's return reasons automatically
- `--inherits-return-reasons=false` opts a new category out, giving it its own (initially empty) list

**`add_return_reasons_to_categories` handles inheritance**
- Adding a reason to an inheriting category materializes it: the inherited reasons are dropped, and the category gets the new reason(s) plus the global reasons appended
- After adding reasons, descendants that inherit are re-resolved so they reflect the updated ancestor

**Docs site**
- The return reasons panel title shows "(inherited)" when a category inherits its reasons

**Schema updated**
- `categories_schema.cue` now allows `return_reasons` to be either a list of strings or `"inherit"`

**Data files updated**
- Categories that share their ancestor's return reasons are now marked `return_reasons: inherit` (2,237 categories); the rest keep their explicit lists
- Includes corrections for some categories that were inheriting their ancestor's return reasons but should define their own. These now carry an explicit list instead of `inherit`

### DIFF
--- a/dev/lib/product_taxonomy/commands/add_return_reasons_to_categories_command.rb
+++ b/dev/lib/product_taxonomy/commands/add_return_reasons_to_categories_command.rb
@@ -25,7 +25,7 @@ module ProductTaxonomy
       @categories.each do |category|
         # A category with no reasons of its own yet — empty, or inheriting from an ancestor — is defined from
         # scratch, so it also gets the global reasons appended (ordered at the end by sort_return_reasons!).
-        from_scratch = category.inherits_return_reasons || category.return_reasons.empty?
+        from_scratch = category.inherits_return_reasons || category.defined_return_reasons.empty?
 
         if category.inherits_return_reasons
           logger.info("Category `#{category.name}` inherited its return reasons - replacing them with the new return reason(s) and the global return reasons")
@@ -49,7 +49,9 @@ module ProductTaxonomy
       # Re-resolve inheritance for descendants so inheriting children reflect their ancestors' updated reasons in
       # the generated artifacts/docs (which read the resolved list, not the literal `inherit`).
       @categories.each do |category|
-        category.descendants.each(&:resolve_inherited_return_reasons)
+        category.descendants.each do |descendant|
+          descendant.resolve_return_reasons if descendant.inherits_return_reasons
+        end
       end
 
       logger.info("Added #{@return_reasons.size} return reason(s) to #{@categories.size} categories")
@@ -57,7 +59,7 @@ module ProductTaxonomy
 
     def append_global_return_reasons!(category)
       ReturnReason.global.each do |return_reason|
-        next if category.return_reasons.include?(return_reason)
+        next if category.defined_return_reasons.include?(return_reason)
 
         category.add_return_reason(return_reason)
       end
@@ -65,7 +67,7 @@ module ProductTaxonomy
 
     # Move the global reasons to the end in their defined order, preserving the order of all other reasons.
     def sort_return_reasons!(category)
-      global, regular = category.return_reasons.partition do |return_reason|
+      global, regular = category.defined_return_reasons.partition do |return_reason|
         ReturnReason::GLOBAL_FRIENDLY_IDS.include?(return_reason.friendly_id)
       end
 
@@ -73,7 +75,8 @@ module ProductTaxonomy
         global.find { |return_reason| return_reason.friendly_id == friendly_id }
       end
 
-      category.return_reasons.replace(regular + sorted_global)
+      category.defined_return_reasons.replace(regular + sorted_global)
+      category.resolve_return_reasons
     end
 
     def update_data_files!
--- a/dev/lib/product_taxonomy/commands/add_return_reasons_to_categories_command.rb
+++ b/dev/lib/product_taxonomy/commands/add_return_reasons_to_categories_command.rb
@@ -2,18 +2,6 @@
 
 module ProductTaxonomy
   class AddReturnReasonsToCategoriesCommand < Command
-    UNKNOWN_RETURN_REASON_FRIENDLY_ID = "unknown"
-    OTHER_RETURN_REASON_FRIENDLY_ID = "other_reason"
-    # Global reasons are always ordered at the end of a category's list, in this exact order.
-    GLOBAL_RETURN_REASON_FRIENDLY_IDS = [
-      "changed_my_mind",
-      "item_not_as_described",
-      "received_the_wrong_item",
-      "damaged_or_defective",
-      UNKNOWN_RETURN_REASON_FRIENDLY_ID,
-      OTHER_RETURN_REASON_FRIENDLY_ID,
-    ].freeze
-
     def initialize(options)
       super
       load_taxonomy
@@ -68,24 +56,20 @@ module ProductTaxonomy
     end
 
     def append_global_return_reasons!(category)
-      global_return_reasons.each do |return_reason|
+      ReturnReason.global.each do |return_reason|
         next if category.return_reasons.include?(return_reason)
 
         category.add_return_reason(return_reason)
       end
     end
 
-    def global_return_reasons
-      @global_return_reasons ||= GLOBAL_RETURN_REASON_FRIENDLY_IDS.map { |friendly_id| ReturnReason.find_by!(friendly_id:) }
-    end
-
     # Move the global reasons to the end in their defined order, preserving the order of all other reasons.
     def sort_return_reasons!(category)
       global, regular = category.return_reasons.partition do |return_reason|
-        GLOBAL_RETURN_REASON_FRIENDLY_IDS.include?(return_reason.friendly_id)
+        ReturnReason::GLOBAL_FRIENDLY_IDS.include?(return_reason.friendly_id)
       end
 
-      sorted_global = GLOBAL_RETURN_REASON_FRIENDLY_IDS.filter_map do |friendly_id|
+      sorted_global = ReturnReason::GLOBAL_FRIENDLY_IDS.filter_map do |friendly_id|
         global.find { |return_reason| return_reason.friendly_id == friendly_id }
       end
 
--- a/dev/lib/product_taxonomy/commands/add_return_reasons_to_categories_command.rb
+++ b/dev/lib/product_taxonomy/commands/add_return_reasons_to_categories_command.rb
@@ -35,8 +35,12 @@ module ProductTaxonomy
       @categories = @categories.flat_map(&:descendants_and_self) if @include_descendants
 
       @categories.each do |category|
+        # A category with no reasons of its own yet — empty, or inheriting from an ancestor — is defined from
+        # scratch, so it also gets the global reasons appended (ordered at the end by sort_return_reasons!).
+        from_scratch = category.inherits_return_reasons || category.return_reasons.empty?
+
         if category.inherits_return_reasons
-          logger.info("Category `#{category.name}` inherited its return reasons - replacing them with the new return reason(s)")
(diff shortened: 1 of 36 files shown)


rachel-carvalho:
* **#983** <a href="https://app.graphite.com/github/pr/Shopify/product-taxonomy/983?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/product-taxonomy/983?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#982** <a href="https://app.graphite.com/github/pr/Shopify/product-taxonomy/982?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

danielpgross (changes requested):
The `regenerate_inherited_return_reasons` command's behaviour of modifying the `data/` files departs from existing conventions in the repo, because it uses definitions in `data/` to derive _other_ definitions in `data/`, then validates that they're in sync using CI. That pattern seems like it could be confusing to new contributors.

Rather than using `inherits_return_reasons` and a derivation into `data/` definitions, could we instead:
- Leave the derivation out of `data/`, doing it instead when we generate the distribution artifacts
- Use the convention below for defining inheritance vs. overrides of return reason definitions: `inherit` means inherit, `[]` means no category-specific reasons (global only), otherwise manually override the definition for this category

```yaml
# Inherit from the nearest defining ancestor
return_reasons: inherit

# Explicitly define no category-specific reasons
# Global reasons would still be added to the generated artifact.
return_reasons: []

# Explicit category-specific reasons
return_reasons:
  - damaged
  - stale
```

The generated distribution would continue to contain each category’s complete effective list, so consumers wouldn't need to implement inheritance themselves. An ancestor edit would automatically appear in inheriting descendants the next time the distribution is generated.

rachel-carvalho:
@dgross I like this suggestion :+1:

ricardotejedorsanz:
I also like the `return_reasons: inherit` contract.

I tested locally and there are two more things I think we should address before approving, plus one nice-to-have:

1. `add_return_reasons_to_categories` needs to account for inheritance. An inheriting category gets a copied effective list, but adding a reason only mutates that in-memory list. Since the data serializer still writes `return_reasons: inherit`, the added reason is discarded on dump/reload. Similarly, changing a defining ancestor leaves already-resolved descendants stale while docs are generated in the same process.

   Could we:
   - Promote an inheriting category to an explicit definition when adding a reason directly to it.
   - Re-resolve inheriting descendants after mutations and before generating artifacts.
   - Add tests confirming both changes persist in generated output and after reload.

2. Nearest-ancestor resolution looks correct: it skips inheriting parents, uses the closest explicit definition, and treats `[]` as an intentional override. However, a category with no defining ancestor, such as a root with `return_reasons: inherit`, silently resolves to an empty list. Could we validate that every inheriting category has a non-inheriting ancestor and fail clearly otherwise?

3. Nice to have: expose `inherits_return_reasons` in the generated JSON alongside the fully materialized `return_reasons` list. Consumers would not need to resolve inheritance, but they could still tell whether the reasons were inherited or explicitly defined.

rachel-carvalho:
> 1. add_return_reasons_to_categories needs to account for inheritance. A

@ricardotejedorsanz Thanks for catching this, I agree that's what it should behave like, I'll make the changes.

> 2. (...) a category with no defining ancestor, such as a root with return_reasons: inherit

I believe a root category shouldn't be able to specify `inherit`, what would it inherit from, right? I'll add a validation to avoid this altogether and it should address the issue of undefined inherited reasons.

> 3. expose inherits_return_reasons in the generated JSON

I can do that too 👍

rachel-carvalho:
@ricardotejedorsanz I made the changes above and also added the `inherited` info to the browser as well:

image

ricardotejedorsanz (approved):
Thanks, @rachel-carvalho, I tested locally and see all changes fixed, including the decision for any manual update to return reasons to completely override the inherited reasons and only keep manual + global. The change to the explorer also looks great.

dev/lib/product_taxonomy/commands/add_return_reasons_to_categories_command.rb
```diff
@@ -4,7 +4,12 @@ module ProductTaxonomy
...
     OTHER_RETURN_REASON_FRIENDLY_ID = "other_reason"
-    SPECIAL_RETURN_REASON_FRIENDLY_IDS = [
+    # Global reasons are always ordered at the end of a category's list, in this exact order.
+    GLOBAL_RETURN_REASON_FRIENDLY_IDS = [
+      "changed_my_mind",
+      "item_not_as_described",
+      "received_the_wrong_item",
+      "damaged_or_defective",
       UNKNOWN_RETURN_REASON_FRIENDLY_ID,
       OTHER_RETURN_REASON_FRIENDLY_ID,
```
danielpgross: nit: These should either be all constants or all strings, not a mix of both.

docs/_layouts/categories.html
```diff
@@ -26,6 +26,7 @@ <h3>Level {{ depth | plus: 1 }}</h3>
                 data-ancestor-ids="{{ category.ancestor_ids }}"
                 data-attribute-handles="{{ category.attribute_handles }}"
                 data-return-reason-handles="{{ category.return_reason_handles }}"
+                data-inherits-return-reasons="{{ category.inherits_return_reasons }}"
```
danielpgross: Related to the other comment -- we can remove this unless it really adds value

dev/lib/product_taxonomy/models/category.rb
```diff
@@ -84,24 +94,27 @@ def add_children(type:, item:, parent:)
...
     # @param parent [Category] The parent category of the category.
     def initialize(id:, name:, attributes: [], return_reasons: [], parent: nil)
       @id = id
       @name = name
       @children = []
       @secondary_children = []
       @attributes = attributes
-      @return_reasons = return_reasons
+      @inherits_return_reasons = return_reasons == :inherit
+      @return_reasons = @inherits_return_reasons ? [] : return_reasons
```
danielpgross: In [this comment](https://github.com/Shopify/product-taxonomy/pull/983#pullrequestreview-4962289909), I was suggesting that we append the global set of return reasons when an empty set of return reasons is defined. Instead, this just keeps the empty set. Which of those behaviours do we want?

rachel-carvalho: This is just assigning a value to `@return_reasons` when inheriting during instantiation before we call `resolve_inherited_return_reasons` during load, which will copy the ancestor's reasons.

https://github.com/Shopify/product-taxonomy/blob/1f929d251c32b6f9e9d9ec69091a95a0a612a50f/dev/lib/product_taxonomy/models/category.rb#L47

https://github.com/Shopify/product-taxonomy/blob/1f929d251c32b6f9e9d9ec69091a95a0a612a50f/dev/lib/product_taxonomy/models/category.rb#L169-L174

I believe you're referring to not keeping empty reasons during add right?  That's the case in the `add` command:

https://github.com/Shopify/product-taxonomy/blob/1f929d251c32b6f9e9d9ec69091a95a0a612a50f/dev/lib/product_taxonomy/commands/add_return_reasons_to_categories_command.rb#L56

Here `from_scratch` is when the reasons were previously empty or when the category was inheriting from an ancestor and will not anymore.

danielpgross: Right, I'm asking about the case during load where `return_reasons: []` is defined explicitly. In that case, there's no inheritance and ancestors won't be resolved. Do we still want the global set of return reasons to be appended to the empty array so it appears in the generated `dist` files?

rachel-carvalho: Thanks for pointing this out. I confirmed the intended behavior with our team and we do want to fallback to global reasons here, so it makes sense to make sure dist and any runtime query results in global reasons for an empty array.

I ended up adding this as part of load and not initialize to ensure that `ReturnReason`s are already loaded, so that we can assign the global reason instances.

https://github.com/Shopify/product-taxonomy/blob/101411b6f62fc2ce47e64a6c9598e00d7bfa17b6/dev/lib/product_taxonomy/models/category.rb#L46-L49

danielpgross: That makes sense. One concern with this approach, though, is that `DataSerializer` will end up writing back the global return reasons instead of keeping the hardcoded `[]`. If another global return reason gets added later, that category won't get it added automatically after that point.

It also leaves categories created after the initial load in an inconsistent state, ex. `AddCategoryCommand` with `inherits_return_reasons: false` creates an empty list but never calls `apply_default_return_reasons`.

Would it make sense to keep the defined value separate from the effective/resolved reasons? The data serializer could keep `[]`, while dist, docs, and runtime callers get `ReturnReason.global`. That would also keep categories created after load consistent without requiring another pass to resolve the global reasons.

sophiedeziel: Implemented an approach based on "Would it make sense to keep the defined value separate from the effective/resolved reasons? "

https://github.com/Shopify/product-taxonomy/pull/983/changes/0a0a262b0b20c2300c5c2ecb383c6e6084d93997

dev/lib/product_taxonomy/models/category.rb
```diff
@@ -43,10 +43,9 @@ def load_from_source(source_data)
         end
         @verticals.sort_by!(&:name)
 
-        # Fourth pass: fall back to the global reasons where a category defines none, then resolve inheritance so
-        # that inheriting categories copy their ancestor's final list.
-        Category.all.each(&:apply_default_return_reasons)
-        Category.all.each(&:resolve_inherited_return_reasons)
+      # Fourth pass: derive each category's effective return reasons — inherited from the closest defining ancestor,
+      # falling back to the global reasons when nothing is defined.
```
danielpgross: nit: fix indentation

danielpgross (approved):
Nice, LGTM 👍

PR #984, opened 2026-08-19
Add a Disclosures tab to the docs release explorer

ruba-ruba:
## TL;DR

Add a **Disclosures** tab to the taxonomy docs site, so the disclosure axis is visible at https://shopify.github.io/product-taxonomy/releases/unstable/disclosures/.

## Screenshots

**A grouping node.** The left box holds the name, the GID, and the names of the child nodes. A click on a child node opens that node.

grouping-view

**A leaf node.** The Details box holds the copy, the display rules, and the public id of the parent. A click on the parent opens the grouping above.

leaf-view

## Why

Shopify/product-taxonomy#956 landed the disclosure data, the model, and the loader. Shopify/product-taxonomy#986 landed the dist serializers. Neither change made the axis visible on the docs site, so partners and internal teams cannot browse disclosures the way they browse categories, attributes, and return reasons. This is the last piece that makes the disclosure axis public, and it unblocks epic shop/issues-product-foundations#4930.

## What

- **Docs serializers:** `Disclosure::Docs::BaseSerializer` writes `docs/_data/<version>/disclosures.yml`, and `Disclosure::Docs::SearchSerializer` writes the search index. Both emit a flat list, because the layout reads the nodes directly.
- **Command:** `GenerateDocsCommand` writes both files, and writes `disclosures.html` in each release folder from `_disclosures_template.html`.
- **Page:** a `disclosures` layout, a nav entry in `_includes/header.html`, and `docs/assets/js/disclosure_release.js` for the client-side filter.
- **Search:** `docs/_plugins/copy_search.rb` copies `disclosure_search_index.json`. Without that entry the page search gets a 404.
- **Unstable release page:** `docs/_releases/unstable/disclosures.html` is committed by hand, because the command skips the release folder for the `unstable` target.

The right box holds the public id, the description, the parent, the jurisdictions, the legal citation, the symbol, the display requirements, the source, the title, and the content. The content comes last, because it is the longest value. The content is markdown, so the layout renders it through `markdownify`.

The Parent row holds the public id of the parent, not the name. The name of a node and the public id of the same node look very different, so the id keeps the row easy to match against the Public id row above it.

A click on a child node, or on the parent, shows that node. The rows are not links, because the page keeps one document and swaps the visible node, as the categories tab does with `.category-node`. Every row takes the focus, so the keyboard reaches it too.

The page leaves out four fields that the dist serializer publishes: `internal_label`, `display_preferences`, `disclosure_attributes`, and `disclosure_attribute_values`. The first one is internal. The other three need a nested row shape, so they stay for a later change.

## Verification

- `rake test:unit` — 505 runs, 0 failures.
- `rake test:integration` — 13 runs, 0 failures.
- RuboCop clean on the changed files.
- A local Jekyll build renders `/releases/unstable/disclosures/` with 10 nodes: 2 groupings and 8 leaves. The build also copies the search index.
- A browser check confirms both directions. A click on the parent row shows the grouping. A click on a child row shows that leaf. The `Enter` key on a focused row does the same.

After merge, the `deploy_github_pages` workflow runs `bin/product_taxonomy docs` on `main`, which regenerates `docs/_data/unstable/`. The page then goes live on the `unstable` release without a wait for the next release cut.

## Known cosmetic issue

The Disclosures nav link renders on every release page, and the older releases have no disclosures page. The link is dead there. Return Reasons has the same flaw today, so this change does not make the behavior worse.

### DIFF
--- a/dev/lib/product_taxonomy/models/serializers/disclosure/docs/base_serializer.rb
+++ b/dev/lib/product_taxonomy/models/serializers/disclosure/docs/base_serializer.rb
@@ -13,25 +13,21 @@ module ProductTaxonomy
             # @param [Disclosure] disclosure
             # @return [Hash]
             def serialize(disclosure)
-              parent = disclosure.parent
               {
                 "id" => disclosure.gid,
                 "public_id" => disclosure.public_id,
                 "name" => disclosure.name,
                 "description" => disclosure.description,
-                "kind" => kind(disclosure),
-                "parent_public_id" => parent&.public_id,
-                "parent_name" => parent&.name,
+                "parent_public_id" => disclosure.parent_public_id,
                 "jurisdictions" => disclosure.jurisdictions,
                 "legal_citation" => disclosure.legal_citation,
+                "title" => disclosure.title,
+                "content" => disclosure.content,
+                "source" => disclosure.source,
+                "symbol" => disclosure.symbol,
+                "display_requirements" => disclosure.display_requirements,
+                "children" => disclosure.children.map { { "public_id" => _1.public_id, "name" => _1.name } },
               }
-            end
-
-            private
-
-            # Every leaf must carry display preferences, so a leaf is always displayed.
-            def kind(disclosure)
-              disclosure.leaf? ? "display" : "grouping"
             end
           end
         end
--- a/dev/lib/product_taxonomy/models/serializers/disclosure/docs/base_serializer.rb
+++ b/dev/lib/product_taxonomy/models/serializers/disclosure/docs/base_serializer.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+module ProductTaxonomy
+  module Serializers
+    module Disclosure
+      module Docs
+        module BaseSerializer
+          class << self
+            def serialize_all
+              ProductTaxonomy::Disclosure.all.map { serialize(_1) }
+            end
+
+            # @param [Disclosure] disclosure
+            # @return [Hash]
+            def serialize(disclosure)
+              parent = disclosure.parent
+              {
+                "id" => disclosure.gid,
+                "public_id" => disclosure.public_id,
+                "name" => disclosure.name,
+                "description" => disclosure.description,
+                "kind" => kind(disclosure),
+                "parent_public_id" => parent&.public_id,
+                "parent_name" => parent&.name,
+                "jurisdictions" => disclosure.jurisdictions,
+                "legal_citation" => disclosure.legal_citation,
+              }
+            end
+
+            private
+
+            # Every leaf must carry display preferences, so a leaf is always displayed.
+            def kind(disclosure)
+              disclosure.leaf? ? "display" : "grouping"
+            end
+          end
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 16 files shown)


jmignac (approved):

dev/lib/product_taxonomy/models/serializers/disclosure/docs/base_serializer.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module ProductTaxonomy
+  module Serializers
+    module Disclosure
+      module Docs
+        module BaseSerializer
```
danielpgross: I don't think `BaseSerializer` is a good name, it sounds like this serializer is supposed to be extended/subclassed. How about `YamlSerializer`?

docs/_layouts/disclosures.html
```diff
@@ -0,0 +1,97 @@
...
+{% include header.html %}
+<div class="attribute-search-wrapper">
+  <div class="search-container">
+    <label for="disclosure-search" class="search-container__label">Disclosure Search:</label>
+    <input class="search-container__input" type="text" id="disclosure-search" placeholder="Loading…">
+    <ul class="search-container__ul" id="disclosure-search-results"></ul>
+  </div>
+</div>
+{% for disclosure in site.data[page.target].disclosures %}
+<div class="return-reason-container hidden" data-public-id="{{ disclosure.public_id }}">
```
danielpgross: We probably shouldn't reuse the `return-reason-container` class here, we could either come up with a new name for disclosures or turn `return-reason-container` into a generic container. 

Same for `return-reasons-wrapper` and similar below

docs/_includes/header.html
```diff
@@ -6,6 +6,7 @@ <h1>Shopify Standard Product Taxonomy ({{ page.title }})</h1>
     <li class="header-links__li"><a href="{{ site.baseurl }}/releases/{{ page.target }}/attributes">Attributes</a></li>
     <li class="header-links__li"><a href="{{ site.baseurl }}/releases/{{ page.target }}/values">Values</a></li>
     <li class="header-links__li"><a href="{{ site.baseurl }}/releases/{{ page.target }}/return_reasons">Return Reasons</a></li>
+    <li class="header-links__li"><a href="{{ site.baseurl }}/releases/{{ page.target }}/disclosures">Disclosures</a></li>
```
danielpgross, line 14: Should we make this render conditionally depending on whether disclosures are available for that release? Otherwise it will render for older releases that don't include disclosures.

dev/lib/product_taxonomy/models/serializers/disclosure/docs/base_serializer.rb
ruba-ruba: Renamed to YamlSerializer

docs/_layouts/disclosures.html
ruba-ruba: changed to disclosure-container

docs/_includes/header.html
ruba-ruba: fixed for Values, Return Reasons and Disclosures. Values and Return Reasons had the same on older releases.

danielpgross (approved):

PR #985, opened 2026-08-19
Add numeric ids to disclosures

ruba-ruba:
## TL;DR

Give each disclosure a stable numeric `id`, and build the global ID from it.

## Why

Shopify Core keys a disclosure by a numeric id. It reads `gid.model_id.to_i` and
rejects a slug, so a slug global ID does not work through the Admin API. This
repo held only `public_id`, so the disclosure axis was not usable from the API.

An internal copy of this data is the live source today. The plan moves the source
to this repo. The ids here match the internal copy, so the global IDs that
merchants already store stay correct after that change.

## What

Data (`data/disclosures.yml`):

- Add a stable numeric `id` to each of the 10 nodes. The values match the
  internal copy.

The hierarchy keeps the slug `parent_public_id`.

Model (`ProductTaxonomy::Disclosure`):

- Build the global ID from `id`.
- Validate that `id` is present, is an integer, and is unique.
- Index disclosures by `id`, so `find_by(id:)` works.

## Product impact

Merchants who set a product disclosure through the Admin API benefit. A slug
global ID fails today, which blocks the disclosure work for product safety and
compliance requirements.

## Verification

- `rake test:unit`: 480 runs, 0 failures.
- `rake test:integration`: 12 runs, 0 failures. The integration suite loads the
  real data file and validates every node.
- RuboCop reports no offenses.
- A new integration test fails if a node loses its `id`, or if two nodes share an
  `id`.
- After the merge, confirm that the downstream table keeps the same ids, and that
  `gid://shopify/TaxonomyDisclosure/{id}` still resolves for the
  `shopify--disclosure-*` metaobjects.

### DIFF
--- a/data/disclosures.yml
+++ b/data/disclosures.yml
@@ -5,9 +5,9 @@
 # warnings, chemical exposure notices, etc.) that a merchant may need to show
 # for a product in a given jurisdiction.
 #
-# Row-shaped: each entry maps 1:1 to a disclosure node, keyed by a stable
-# `public_id` slug — never rename an existing `public_id`, as downstream systems
-# reference it by global ID.
+# Row-shaped: each entry maps 1:1 to a disclosure node. `id` is the numeric
+# identity that global IDs use, and `public_id` is a stable slug. Never change
+# an existing `id` or `public_id`, as downstream systems reference both.
 #
 # Hierarchy is expressed via `parent_public_id`. A node with no `parent_public_id`
 # is a grouping/root node; a node that no other node points at is a leaf.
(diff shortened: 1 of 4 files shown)


data/disclosures.yml
```diff
@@ -46,14 +50,18 @@
   display_preferences:
     surfaces:
     - product_page
+  display_preference:
+    surfaces:
+    - product_page
```
danielpgross: looks like a duplicate of the above, added by accident?

data/disclosures.yml
```diff
@@ -66,14 +74,18 @@
...
```
danielpgross: mistake?

danielpgross (dismissed):
`display_preference` is a straight duplicate of `display_preferences` -- I don't understand why we would need both, even if there's some legacy support we're trying to achieve. We should either rename `display_preferences` to `display_preference`, or keep it as `display_preferences` and let downstream consumers do the necessary rename or map.

jmignac (approved):

danielpgross (approved):

PR #995, opened 2026-08-25
Add the disclosures docs data for the 2026-08 release

ruba-ruba:
## TL;DR

The 2026-08 release ships disclosures in the distribution, but the docs explorer has no disclosures page for that release. This change adds the two data files and the page.

## Why

PR #984 added the code that generates the disclosures docs. The 2026-08 release was cut before that merge, so the release folder has no disclosure files. Nothing regenerates the docs of a release after the cut, so the gap stays until someone adds the files. PR #954 did the same backfill for return reasons in 2026-02 and 2026-05.

Impact: a merchant or a partner who reads the 2026-08 release in the explorer cannot see the disclosures that the release contains. The distribution and the docs must agree.

## What

- `docs/_data/2026-08/disclosures.yml`
- `docs/_data/2026-08/disclosure_search_index.json`
- `docs/_releases/2026-08/disclosures.html`

The data comes from the `v2026-08` tag, not from `main`. Two CPSC eFiling disclosures merged after the release cut, so the 2026-08 page must not show them. The page lists the 10 disclosures of the release.

Note for the reviewer: do not reproduce this with `docs --version=2026-08` in a clean checkout. The command also rebuilds `mappings.yml` from the distribution on disk, which relabels the release as `shopify/2026-11-unstable`. This change writes only the three files above.

## Verification

A local Jekyll server confirms the result:

- `/releases/2026-08/disclosures/` returns 200 and lists 10 disclosures.
- The page has no CPSC eFiling entries.
- `/releases/2026-08/disclosure_search_index.json` returns 200.
- The header shows the Disclosures link for 2026-08 and hides it for the older releases that have no disclosure data.
- The file set of `docs/_data/2026-08` now matches `docs/_data/unstable`.

The three files are byte-identical to the output of `bin/product_taxonomy docs --version=2026-08`, run on `main` with the `data` directory of the `v2026-08` tag.


KaiKitJeffreyChan (approved):
Nice LGTM!

PR #1009, opened 2026-09-15
Add EU toy safety disclosures to the public taxonomy

ruba-ruba:
Add 15 EU Toy Safety Directive disclosures so the next taxonomy release includes standard warnings for toys unsuitable for children under 36 months.

- Add one group and 15 leaves with stable IDs and slugs.
- Keep warning content, names, internal labels, 30 EU/EEA jurisdictions, product-page display, and no default icon unchanged.
- Regenerate English localization and disclosure outputs for 31 locales. Generated files also include two existing CPSC entries absent from the previous outputs.

**Validation:** 542 unit tests and 13 integration tests pass. Data checks verify the approved copy, metadata, and all locale exports. CUE validation remains blocked by a missing CLI; this repository has no dev setup. Local Binks review is unavailable because its package fetch returns HTTP 400.

**Release verification:** Confirm IDs 23–38 appear in the published disclosure assets. Check Norway exports as `"NO"`, symbols are null, and surfaces contain only `product_page`.


data/disclosures.yml
```diff
@@ -252,3 +252,768 @@
...
+  - IS
+  - IE
+  - IT
+  - LV
+  - LI
+  - LT
+  - LU
+  - MT
+  - NL
+  - "NO"
```
darwinShopify, line 300: Is this supposed to be in quotes?

ruba-ruba: I need to wrap it otherwise it evaluates to `false` when parsing the file

darwinShopify (approved):

JeffrayZhang (approved):

